PR #1, opened 2019-03-15
node 10 required in engines

madflow:
I began to debug the error 
```
TypeError: Promise.all(...).finally is not a function
``` 
after a worker shutdown, when I realized that Promise.finally had been introduce in Node 10. This PR documents this in the engines field.


benjie:
Released in `0.0.1-alpha.2`

Issue #8, opened 2019-04-08
Please export `getTasks`


jamesallain:
So that graphile-worker can be used programmatically.

benjie:
Current workaround it to do something like this:

```js
import { start as startWorkers } from "graphile-worker";
import getTasks from "graphile-worker/dist/getTasks";
getTasks(`${__dirname}/tasks`).then(
  ({ tasks }) => {
    startWorkers(tasks, pgPool, { workerCount: 5 });
  },
  err => {
    console.error(err);
    process.exit(1);
  }
);
```

but we should definitely avoid importing things from `dist`!

1. we should expose `getTasks`
2. we should expose a convenience wrapper that does all this boilerplate. I've decided this wrapper should be called `start`, so we need to rename the existing `start` to `makeWorkerPool`

Aim:

```js
import { start } from 'graphile-worker';

const managedPool = start({
  tasksPath: `${__dirname}/tasks`,
  connectionString: process.env.DATABASE_URL,
  // or pgPool: new Pool(...)
  workerCount: 5,
});

// ...

await managedPool.stop();
```

I'm thinking there should be a `ManagedWorkerPool` that's one level higher than the `WorkerPool` (it's responsible for loading the tasks too).

benjie:
@geekuillaume is interested in taking this on 👍

benjie:
https://github.com/graphile/worker/blob/master/src/interfaces.ts

```
export interface Options extends WorkerPoolOptions {
  connectionString?: string;
  pgPool?: Pool;
  tasksPath?: string;
}
```

PR #9, opened 2019-04-10
Programmatic API

geekuillaume:
This change expose a new `initWorker` that allows a NodeJS program to launch workers from its code without using the CLI.

Related to #8


benjie:
Thanks for the PR @geekuillaume; I've refactored it and merged it in #10 👍

Issue #7, opened 2019-04-06
Add examples of using worker to send tasks to other queues


benjie:
- [x] Faktory
- [ ] SQS
- [ ] RabbitMQ
- [x] GCP Cloud Tasks

madflow:
Where would these examples go? 

* [ ] Wiki
* [ ] Readme
* [ ] /examples

I have used a similar approach to export database events to https://github.com/contribsys/faktory in the past. So this could be an example...

benjie:
Lets go with `/examples` for now 👍

garcianavalon:
I would send an example on how to delegate to GCP Cloud Tasks

benjie:
Awesome :raised_hands:  Follow the pattern set out by [the faktory example](https://github.com/graphile/worker/tree/master/examples/worker-faktory-exporter) please :+1:

garcianavalon:
PR open @benjie

keithlayne:
I can do SQS if nobody has before I get to it.  Wrote SQS->worker code today in my project, will do the other direction soon.

benjie:
Yes please @keithlayne :raised_hands:

benjie:
(Also if you want to show an SQS -> worker example too, that'd also be welcome!)

benjie:
Closing this; I'm happy with what we have.

PR #11, opened 2019-04-13
Add example for Faktory #7

madflow:
* I experimented with exporting Postgres `NOTIFY` events to  [Faktory](https://github.com/contribsys/faktory) because I could not find a Postgres job queue that would meet my needs. So this is a rip-off from that work.
* I was not quite sure I a task function should throw on an error or return something to fail the job...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,4 @@ yarn-debug.log*
 yarn-error.log*
 .vscode
 node_modules/
-tasks/
+/tasks/
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,0 +1,6 @@
+# Examples
+
+This directory contains examples of use cases for `graphile-worker`.
+
+* [worker-faktory-exporter](./worker-faktory-exporter): A way of exporting jobs to the [Faktory](https://github.com/contribsys/faktory) work server. 
+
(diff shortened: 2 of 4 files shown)


benjie:
LGTM 👍 

And yes, the tasks should throw on error.

benjie:
Thanks for this @madflow; this is going to be a great place to point people to when they ask about similar tasks.

benjie:
Does Faktory have a unique identifier for jobs? Was wondering if it was worth using that to enforce idempotency - at the moment if `await faktoryClient.close();` throws then the job will be re-queued and may be added to Faktory multiple times.

madflow:
You can provide the `jid` in the job payload, which is a unique identifier. You are right - this would be worth using.  I will investigate :running_man: .

benjie:
You can het our job id from `helpers.job.id` 👍

madflow:
It turns out, that unique jobs are a "Pro" feature: https://github.com/contribsys/faktory/wiki/Pro-Unique-Jobs

It is possible to pass the jid in the job payload - but you can do this multiple times. There is no documented way I am aware of to check, if the current jid is already being processed by Faktory.

benjie:
👍

PR #20, opened 2019-07-24
Resolves #17 - Gives DATABASE_URL env var a chance to shine

### LINKED ISSUE #17: When using as a library - connectionString or pgPool are mandatory
The documentation states, that either connectionString or pgPool must be set - or the env var DATABASE_URL is available.

However: A set DATABASE_URL is ignored.

See https://github.com/graphile/worker/blob/40d3df62e81e8350488d4c0c33b3a863dd81d91b/src/runner.ts#L33

bbarr:
Might be a little more work to do here, plus test coverage for setting options, but this quickly let's us use process.env.DATABASE_URL.

### DIFF
--- a/src/runner.ts
+++ b/src/runner.ts
@@ -29,7 +29,7 @@ const processOptions = async (options: RunnerOptions) => {
     }
 
     assert(
-      !!options.pgPool !== !!options.connectionString,
+      !(!!options.pgPool && !!options.connectionString),
       "Exactly one of either pgPool or connectionString should be set"
     );
     let pgPool: Pool;


src/runner.ts
```diff
@@ -29,7 +29,7 @@ const processOptions = async (options: RunnerOptions) => {
     }
 
     assert(
-      !!options.pgPool !== !!options.connectionString,
+      !(!!options.pgPool && !!options.connectionString),
```
benjie: To avoid the double-negatives, could we go with something like:

```suggestion
      !options.pgPool || !options.connectionString,
```

Other than that, this looks good to merge :+1:

src/runner.ts
```diff
@@ -29,7 +29,7 @@ const processOptions = async (options: RunnerOptions) => {
...
       "Exactly one of either pgPool or connectionString should be set"
```
benjie:
```suggestion
       "Both pgPool and connectionString are set, at most one of these options should be provided"
```

benjie:
Fixed by #27

PR #21, opened 2019-07-24
Add note about how to run from functions with less-privileged users

bbarr:


README.md
```diff
@@ -363,6 +363,11 @@ You can skip parameters you don't need by using PostgreSQL's named parameter sup
 SELECT graphile_worker.add_job('reminder', run_at := NOW() + INTERVAL '2 days');
 ```
 
+**NOTE:** If `graphile_worker.add_job` is inside a PostgreSQL function
+that will be executed by a user with limited privileges then the function
+must include the `SECURITY DEFINER` flag so that it runs 
+with the same privileges as the more powerful user that defined it.
```
benjie: Excellent start; how about this?

```
**NOTE:** `graphile_worker.add_job(...)` requires database owner privileges
to execute. To allow lower-privileged users to call it, wrap it inside a
PostgreSQL function marked as `SECURITY DEFINER` so that it will run
with the same privileges as the more powerful user that defined it. (Be
sure that this function performs any access checks that are necessary.)
```

PR #24, opened 2019-08-22
load tasks written in typescript

mathroc:
makes it possible to run: `ts-node /scripts/node_modules/.bin/graphile-worker`

maybe there's another way of doing so but I'm just beginning with typescript and don't know the ecosystem very well

I tried looking for an env var to detect if we're running with ts-node but didn't find anything

if this is an interesting feature, I can a cli flag to enable this behavior

### DIFF
--- a/src/getTasks.ts
+++ b/src/getTasks.ts
@@ -105,7 +105,7 @@ export default async function getTasks(
     if (watch) {
       watchers.push(
         chokidar
-          .watch(`${taskPath}/*.js`, {
+          .watch(`${taskPath}/*.{j,t}s`, {
             ignoreInitial: true,
           })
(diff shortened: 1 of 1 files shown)


benjie:
I wouldn’t advise this approach in production.

Instead I have a src/tasks folder which I compile to dist/tasks using `tsc -w` in development for watch mode or just `tsc` in production. This minimises the memory usage, and improves startup/restart time.

benjie:
I am planning on adding “adaptors” though, to run scripts in various languages. It’s not a high priority right now though.

mathroc:
ah alright, thanks for the suggestion :)

benjie:
Thanks for raising the issue, I'm sure others will look for this in future :+1:

Issue #22, opened 2019-07-26
PgPools don't have error handlers


benjie:
https://github.com/graphile/worker/blob/a8070aeb3da7de1e0b83ec32568effa83e4d3a99/src/runner.ts#L39-L43

Should have `.on('error', ...)` handlers to prevent node crashing when PostgreSQL restarts.

benjie:
Fixed in #26; thanks @madflow!

PR #26, opened 2019-09-13
Add pgPool 'error' handler

madflow:
* #22 
* The process does not exit anymore in case of a backend restart - but there is a Promise Rejection Warning:

```
1|worker  | (node:64357) UnhandledPromiseRejectionWarning: Error: Client has encountered a connection error and is not queryable
1|worker  |     at process.nextTick (.../node_modules/pg/lib/client.js:521:25)
1|worker  |     at process._tickCallback (internal/process/next_tick.js:61:11)
1|worker  | (node:64357) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
1|worker  | (node:64357) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code
```

* Maybe there is more work to do here...

### DIFF
--- a/src/runner.ts
+++ b/src/runner.ts
@@ -47,6 +47,19 @@ const processOptions = async (options: RunnerOptions) => {
       );
     }
 
+    pgPool.on("error", err => {
+      /*
+       * This handler is required so that client connection errors don't bring
+       * the server down (via `unhandledError`).
+       *
+       * `pg` will automatically terminate the client and remove it from the
+       * pool, so we don't actually need to take any action here, just ensure
+       * that the event listener is registered.
+       */
+      // eslint-disable-next-line no-console
+      console.error("PostgreSQL client generated error: ", err.message);
+    });
+
     const withPgClient = makeWithPgClientFromPool(pgPool);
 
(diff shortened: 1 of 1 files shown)


src/runner.ts
```diff
@@ -47,6 +47,11 @@ const processOptions = async (options: RunnerOptions) => {
       );
     }
 
+    // Prevent node to emit an uncaught error and exit when the backend restarts or goes down
+    pgPool.on('error', err => {
+      console.error('The PostgreSQL client emitted an error: ', err.message);
+    });
+
```
benjie: Thanks for spotting my lack of this; please can you copy the code from here as it has a comment explaining it too:

https://github.com/graphile/postgraphile/blob/d2374174adfb363c20fc068aea255ad45bbbe708/src/postgraphile/postgraphile.ts#L211-L222

Thanks :pray:

PR #30, opened 2019-10-04
Add postgresql 12 to ci grid

madflow:
* Postgres 12 is released :tada: 
* Maybe this is a little bit "over the top" with all the different versions - but I introduced this in a prior PR, so here it goes.


madflow:
Done and rebased.

PR #31, opened 2019-10-04
How to solve issue when extensions aren't in public

archlemon:
This is how I fixed the issue on an old database that has extensions installed in another schema. Not sure if this is really a nice/elegant way of fixing this.

Maybe I should list the used functions, or link to an external file/readme with the code or solution to fix this? I don't really want to pollute the main readme with edge cases.

What's your thought on the best place to put this info?

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,7 @@ PostgreSQL 10+\* and Node 10+\*.
 If your database doesn't already include the `pgcrypto` and `uuid-ossp`
 extensions we'll automatically install them into the public schema for you. If
 you have them installed in a different schema (unlikely) you may face issues.
+Making alias functions in the public schema, should solve this issue.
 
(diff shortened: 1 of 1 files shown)


benjie:
Maybe link to this issue in that text, and then outline the functions you created that were necessary to work around this issue in your initial comment? :point_up: 

Honestly I was hoping no-one would be hit by this; I'm surprised it came about so fast! Are you a PostGraphile user or are you using graphile-worker with something else?

archlemon:
Ah yeah, an issue is probably a better solution.

I'm currently implementing worker, switching from an old (horrid) php solution. It's also using postgraphile, but I think the extensions were added to another schema for some other tools or something. Postgraphile was added later in the projects life.

I'll try get the issue created and linked this weekend/early next week.

Thanks!

Issue #23, opened 2019-08-08
Add logging helper (or override console.log?) that prefixes the worker id/task


benjie:
We already have debug, but one for regular logging would be helpful.

benjie:
Achieved in #32

PR #32, opened 2019-10-09
feat(logs): allow replacing console logging with a custom solution

bmhaskar:
Tried to remove the console.log with default logger.
Made sure that defaultMeta indicating worker-id  is available with every logger.

Ref: https://github.com/graphile/worker/issues/23


src/logger.ts
```diff
@@ -0,0 +1,41 @@
+import * as winston from "winston";
+import { Logger } from "winston";
```
benjie: Prefer `const { Logger } = winston;` to avoid the double import.

src/logger.ts
```diff
@@ -0,0 +1,41 @@
...
+
+const errorFilePath = path.resolve("..", "logs", "error.logs");
+const combinedErrorFilePath = path.resolve("..", "logs", "combined.logs");
+
+const container = new winston.Container();
+
+const winstonOptions = {
+  level: LOGGER_LEVEL,
+  format: winston.format.json(),
+  defaultMeta: { "worker-id": "dfault-graphile-worker" },
```
benjie: Typo; also would it be better to drop the `default-` anyway?

```suggestion
  defaultMeta: { "worker-id": "graphile-worker" },
```

src/logger.ts
```diff
@@ -0,0 +1,41 @@
...
+  transports: [
+    //
+    // - Write to all logs with level `info` and below to `combined.log`
+    // - Write all logs error (and below) to `error.log`.
+    //
+    new winston.transports.File({ filename: errorFilePath, level: "error" }),
+    new winston.transports.File({ filename: combinedErrorFilePath }),
```
benjie: Logging to file should not be enabled by default, please. Can we allow the user to specify this through configuration somehow?

src/logger.ts
benjie: Alternatively 

```suggestion
export const Logger = winston.Logger;
```

to achieve 2 goals with one line of code.

bmhaskar: Logger is not declared as part of winston type, it wont compile.

benjie: That suggests the `import * as winston` is misleading then. I wrote `export const Logger = ` before, but it seems I meant `export type Logger =`

benjie:
I've thought about this a fair amount more and I'm not happy adding dependencies to graphile-worker for this feature because it will increase my maintenance burden due to security updates/etc and I'm perfectly happy with the use of `console` in the worker project (it's intended to be used for batch jobs so the trade-offs of console are not particularly problematic IMO).

I am, however, happy to support a generic logging interface to which you can pass a logger instance (which will default to a built-in console-based logger). We could even potentially pass this on the command line, e.g. `graphile-worker --log-adapter graphile-worker-winston-log-adapter`

Would you be interested in building this interface? If you drop the `callback` from the above specs I'm happy with that being the defined interface for simple compatibility with Winston.

benjie:
I've overhauled this so that there's only a simple function factory you need to implement; e.g.

```js
function logFactory(scope) {
  return (level, message, meta) => {
    console.log(level, message, scope, meta);
  };
}
```

This should make integrating with all sorts of logging providers very easy to deal with. It also gives us the flexibility to change our internal logging class however we need without having to worry about backwards compatibility too much.

What do you think?

bharat-mhaskar-roostify:
I did not realize the IOC you intended earlier, though this is a really good implementation.

benjie:
I didn't realise this was what I wanted until I'd fiddled about with it for a while myself. Thanks for bearing with me!

PR #51, opened 2019-12-01
Make the task payload of type any

testarossaaaaa:
It allows instead of it:
```typescript
function email(payload: unknown, helpers: Helpers) {
  const p = payload as Payload
  ...
}
```
write it:
```typescript
function email(payload: Payload, helpers: Helpers) {
  ...
}
```

### DIFF
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -36,7 +36,7 @@ export interface Helpers {
   addJob: AddJobFunction;
 }
 
-export type Task = (payload: unknown, helpers: Helpers) => void | Promise<void>;
+export type Task = (payload: any, helpers: Helpers) => void | Promise<void>;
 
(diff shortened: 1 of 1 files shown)


benjie:
I think this is unsafe as we cannot guarantee the shape of the job will be like that. Better to have a type guard function or use runtypes or similar to take the unknown and turn it into a safe object.

testarossaaaaa:
It makes sense. But practically it will be a runtime error that should fail the job. Just for our project I added a module.d.ts file with the following content:
```typescript
import { Helpers } from 'graphile-worker'

declare module 'graphile-worker' {
  type Task = (payload: any, helpers: Helpers) => void | Promise<void>
}
```
Maybe it will be useful for someone.

PR #52, opened 2019-12-10
Re-export `defaultLogger` from top level

keithlayne:
Also adds documentation on one potential use case: modifying the default
logger scope.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -3,6 +3,11 @@ export * from "./interfaces";
 
 export { runTaskList, runTaskListOnce } from "./main";
 export { run, runOnce, runMigrations } from "./runner";
-export { Logger, LogFunctionFactory, consoleLogFactory } from "./logger";
+export {
+  Logger,
+  LogFunctionFactory,
+  consoleLogFactory,
+  defaultLogger,
+} from "./logger";
 
 export { getTasks };
--- a/README.md
+++ b/README.md
@@ -265,6 +265,17 @@ main().catch(err => {
 
 ### The logger option
 
+If you want to modify just the scope of the default logger, you can do so simply.  For example:
+
+```js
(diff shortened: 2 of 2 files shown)


src/index.ts
```diff
@@ -3,6 +3,6 @@ export * from "./interfaces";
 
 export { runTaskList, runTaskListOnce } from "./main";
 export { run, runOnce, runMigrations } from "./runner";
-export { Logger, LogFunctionFactory, consoleLogFactory } from "./logger";
+export { Logger, LogFunctionFactory, consoleLogFactory, defaultLogger } from "./logger";
```
keithlayne: Prettier want to expand this line now...don't know if that's in your CI.  I will leave this unless you tell me otherwise or CI fails.

keithlayne: Should have waited.

keithlayne:
This is no good.  My example and intended use case don't even work because of https://github.com/graphile/worker/blob/master/src/worker.ts#L40-L43 .

Let me know if you're open to changing this behavior so we can override `workerId` at the top level.

I know for some people it won't even make sense, and maybe I should just not try to do this.  I'm running only one worker per process.

benjie:
You can already pass a custom `workerId` via options? But yes I'm open to letting you specify a logger and having the default behaviour just be the fallback.

keithlayne:
Ugh.  I see `WorkerOptions` but you can't pass that to `run`:

https://github.com/graphile/worker/blob/master/src/interfaces.ts#L117-L129

Also, `workerId` is not mentioned in the readme except as a parameter to `logFactory`.  I am, as usual, feeling stupid.

keithlayne:
To be clear, looks like the JS will pass that option just fine, but the types don't support it.

benjie:
PR to fis the types and docs welcome; or raise an issue :+1:

PR #54, opened 2019-12-11
fix(types): add `workerId` to `RunnerOptions`

keithlayne:
This PR also adds documentation to the readme for this option.  I'd
appreciate any feedback on that part.

### DIFF
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -121,7 +121,7 @@ export interface WorkerOptions extends WorkerSharedOptions {
   workerId?: string;
 }
 
-export interface WorkerPoolOptions extends WorkerSharedOptions {
+export interface WorkerPoolOptions extends WorkerOptions {
   /**
    * Number of jobs to run concurrently
    */
--- a/README.md
+++ b/README.md
@@ -217,6 +217,7 @@ The following options for these methods are available.
 - `concurrency`: The equivalent of the cli `--jobs` option with the same default value.
 - `pollInterval`: The equivalent of the cli `--poll-interval` option with the same default value.
 - `logger`: To change how log messages are output you may provide a custom logger; see `logger` below
+- `workerId`: Static worker ID used for created workers; should only be used with `concurrency: 1`
 - the database is identified through one of these options:
   - `connectionString`: A PostgreSQL connection string to the database containing the job queue, or
   - `pgPool`: A `pg.Pool` instance to use


src/interfaces.ts
```diff
@@ -121,7 +121,7 @@ export interface WorkerOptions extends WorkerSharedOptions {
   workerId?: string;
 }
 
-export interface WorkerPoolOptions extends WorkerSharedOptions {
+export interface WorkerPoolOptions extends WorkerOptions {
```
benjie: This isn't valid; a worker pool is for managing multiple workers (a pool of workers); each worker needs its own ID, so specifying a workerId at this level would only work for a pool of size 1.

benjie: (I don't want to open this can of worms; can we instead change the relevant place that uses WorkerPoolOptions that you think should use WorkerOptions instead?)

keithlayne: > Going back after writing this, it's kinda all over the place.  As you read, just pick out the parts that make any sense :)

So I was working off the possibility that it was a typo to begin with, and the apparent usage of the options in JS. I realized the issues you mention afterwards.  I don't think there's a place to swap it out, and like I said the options seemed to be destructured in a way consitent with this change. (maybe not? see below)

I had put in the docs "only use with concurrency 1" but all that seems pretty dumb.  I agree this is a bad change.

I can't really think of a satisfying solution other than a config option that generates custom worker IDs.  

Overall, the default behavior serves *my* needs well, so I think probably just reject this change.  You had told me on discord that you can pass `workerId` at the top level, and you can in JS and it would work in a way consistent with this change, which is a little weird.  But I guess the more correct typing would be to revert the above and instead do:

```ts
export interface RunnerOptions extends WorkerOptions, WorkerPoolOptions {...}
```

keithlayne: But...no matter how you expose `workerId` at the top level, it's still problematic like you said because it will break with concurrency > 1.  So I'd actually kinda prefer disallowing it at the top level and passing an optional generator function.  But it's not clear that anyone wants or needs that feature, so I think I still kinda vote to close this unless there's something I'm missing.

benjie:
I think closing this makes sense. But if you want to control the workerId, perhaps we can make it easier to run a single instance (not a pool) of the worker - that way passing the workerId would make sense?

benjie:
Also you're right that we don't prevent `workerId` being specified at the root level; if you wanted to add a PR to throw on that case if `concurrency !== 1` that'd be nice. I'm happy with the TS typings for my own projects.

PR #60, opened 2019-12-19
feat: add support for efficiently scheduling jobs from JS

### LINKED ISSUE #55: addJob - helper function
I'm using this as lightweight event bus. There are a bunch of places in my platform where events need to be produced from random code blocks. The intent of this library appears to be queue messages from db triggers, but in my case there is no db corresponding db write and synthesizing one seems hamfisted. 

Currently, the addJob function is attached to the runner instance. It's be cool if the addJob function was exported separately that either generate the query and the call params or take a db connection and run the insert. The reason for the connection and not a pool is that cloud functions dont do well with pools and people might want to queue messages from webhooks or something.

The main argument for exporting this and having it travel with this library is that if the db schema changes or something, it would also get updated.

If you are open to it, I'll do the work.  Would you accept a PR that exports the addJob function standalone? 

If not, this is no big deal. It is easy to queue with the sproc. Thank you for making this lib. :)

mrmurphy:
This should resolve #55


README.md
```diff
@@ -69,6 +69,32 @@ you're ready, switch to using the `package.json` `"scripts"` entry instead.)
...
+
+#### The Publisher API
+
+Import `runPublisher` from `graphile_worker`, pass in your database configuration options (just like you would with `run` above.) and then pass a new job into `addJob` with a name and a payload!
+
+```js
+const { runPublisher } = require("graphile-worker");
+
+async function main() {
+  const publisher = await runPublisher({ connectionString: "postgres:///" });
```
benjie: Lets make this more generic, so we can use it to expose more helpers in future. And "run" doesn't really feel like the right verb here.

```suggestion
  const utils = new Utils({ connectionString: "postgres:///" });
```

We should also make it clear (if necessary) that you should call `utils.end()` or similar when you're done with it.

I think we should also make an even more direct method for adding a job that wraps this:

```js
export async function addJob(config, ...args) {
  const utils = new Utils(config);
  try {
    return await utils.addJob(...args);
  } finally {
    utils.end();
  }
}
```

Using this directly a lot would be expensive due to the pool acquisition and releasing, but for the odd one-off might be better than passing a `utils` instance around?

mrmurphy: I see where you're coming from if your primary method of adding jobs is through the database, and the application only adds jobs now and then.

From my current architecture, and what I think the average JS server developer would expect, this `publisher` API would be my primary way of adding jobs.

```
    +-------------+ +-----------------+                   
    | Application | | Worker process  |                   
    +-------------+ +-----------------+                   
           |                 |                            
 Publisher |                 |                            
           |                 | Graphile CLI               
           |                 |                            
    +---------------------------------+                   
    |    Postgres                     |                   
    +---------------------------------+                   
               |     |                                    
               |     |                                    
               |     |                                    
               |------                                    
               Queries or triggers                        
```

In other words, I love the simplicity of the CLI, and don't actually intend to use the programmatic API for subscribing to jobs. (Not that the programmatic API isn't useful, but I think the CLI may be the way the majority of people will run the workers.)

Even though putting `addJob` into a `utils` scope would function just the same as what we have in this PR, I'd argue that even the name `utils` makes the approach of adding jobs through Node feel secondary. Do you think there's a compromise we can make about the way things are named, so that bootstrapping a publish-only connection for an application feels like a first-class pattern?

Please let me know if you think I'm missing the boat entirely here, as well 😅

benjie: I don't agree; users'd run something like:

```js
const workerUtils = new WorkerUtils({connectionString});

async function main() {
  // ... do stuff
  workerUtils.addJob(...);
  // ... do more stuff
  workerUtils.addJob(...);
  // ... yet more stuff
}
```

This `new Thing` approach is a common pattern in the Node ecosystem, it feels very natural, and it means you can create the utils in the global scope without having to `await` (and you can easily pass it around your app). It also means that you can ultimately `workerUtils.end()` if you need to clean things up (e.g. for tests).

As time goes on we can add more utils, like `workerUtils.getFailedJobs()`, `workerUtils.getQueueLength("QUEUE NAME HERE")` or `workerUtils.postponeJob(jobId, untilDate)` which I think makes a lot of sense for "utils" but wouldn't really make sense for "publisher".

The reason I'd go with `Utils` or `WorkerUtils` rather than `Worker` is because I don't want people to confusing instanciating it with actually running the worker. The reason I don't want to go with `Publisher` is because it's too specific, and I want to be able to add more helpers over time (as above).

If you can think of a better name that addresses these issues, I'm all ears.

mrmurphy: Okay, that makes sense to me. I’ll update the PR sometime soon here.

mrmurphy:
@benjie does this look like what you were thinking? It took me a lot longer than I thought because I haven't done Typescript in a while. But I think it should be ready to merge now. I added a test for each function.

README.md
```diff
@@ -69,6 +69,56 @@ you're ready, switch to using the `package.json` `"scripts"` entry instead.)
...
 ### Schedule a job:
 
+There are two ways to schedule jobs:
+
+1. From Node with the `WorkerUtils` or `addJob` APIs.
+2. In the database directly with the `add_job` SQL function.
+
+#### The WorkerUtils API
+
+Import `WorkerUtils` from `graphile_worker`, and make a new instance by passing in your database configuration options (just like you would with `run` above). What you get back is an instance of a class with a function called `addJob`, and a function called `end`.
```
benjie:
```suggestion
Import `WorkerUtils` from `graphile-worker`, and make a new instance by passing in your configuration:

- one of these to determine how to connect to the database:
  - `connectionString`: A PostgreSQL connection string to the database containing the job queue, or
  - `pgPool`: A `pg.Pool` instance to use
 
A `WorkerUtils` instance has the following methods:

- `end` - call this to release the `WorkerUtils` instance. It's typically best to use `WorkerUtils` as a singleton, so you often won't need this, but it's useful for tests or processes where you want Node to exit cleanly when it's done.
- `addJob(name: string, payload: JSON, options: TaskOptions)` - a method you can call to enqueue a job, it returns a promise that resolves to the job. `options` accepts the following parameters:
  - `run_at`?
```

- [ ] TODO: document options.

README.md
```diff
@@ -69,6 +69,56 @@ you're ready, switch to using the `package.json` `"scripts"` entry instead.)
...
+
+The `addJob` function takes a name, and a payload, and returns a promise of a job that has been added to the queue. The `end` function closes the connection, don't forget to call it yourself if you're done adding jobs and don't want to keep the pool open anymore!
```
benjie: Now covered above.

```suggestion
```

README.md
```diff
@@ -69,6 +69,56 @@ you're ready, switch to using the `package.json` `"scripts"` entry instead.)
...
+  }
+}
+
+main().catch(err => {
+  console.error(err);
+  process.exit(1);
+});
+```
+
+##### The `addJob` standalone Javascript function
```
benjie:
```suggestion
##### The `addJob` standalone JavaScript function
```

README.md
```diff
@@ -69,6 +69,56 @@ you're ready, switch to using the `package.json` `"scripts"` entry instead.)
...
+
+If you just want the fast and easy way to add a job, and you don't mind the cost of opening a DB connection pool, and then cleaning it up right away _for every job added_, there's the `addJob` convenience function:
```
benjie:
```suggestion
If you just want the fast and easy way to add a job, and you don't mind the cost of opening a DB connection pool, and then cleaning it up right away _for every job added_, there's the `addJob` convenience function. It takes the `WorkerUtils` configuration as the first argument, and then the remaining `WorkerUtils#addJob` arguments afterwards:
```

benjie: Note: I'm trying to be careful to use "configuration" for the WorkerUtils argument, and "options" for addJob to avoid confusion of just using "options" everywhere.

__tests__/worker_utils.addJob.test.ts
```diff
@@ -0,0 +1,42 @@
...
+    // Assert that it has an entry in jobs / job_queues
+    const { rows: jobs } = await pgClient.query(
+      `select * from graphile_worker.jobs`
+    );
+    expect(jobs).toHaveLength(1);
+
+    const task: Task = jest.fn();
+    const taskList = { task };
+    await runTaskListOnce(taskList, pgClient);
+  }));
```
benjie: Please also write a test passing job options such as `runAt`

src/interfaces.ts
```diff
@@ -146,3 +146,19 @@ export interface RunnerOptions extends WorkerPoolOptions {
    */
   pgPool?: Pool;
 }
+
+export interface WorkerUtilsOptions {
```
benjie: We should maybe rename this to "configuration" so that "configuration" and "options" remain distinct. If you can think of better words (or think they should be the other way around) please let me know :)

src/interfaces.ts
```diff
@@ -146,3 +146,19 @@ export interface RunnerOptions extends WorkerPoolOptions {
...
+  /**
+   * A PostgreSQL connection string to the database containing the job queue
+   */
+  connectionString?: string;
+  /**
```
benjie:
```suggestion

  /**
```

src/worker_utils.ts
```diff
@@ -0,0 +1,62 @@
+import { Pool, PoolClient } from "pg";
+import { WorkerUtilsOptions, TaskOptions, Job } from "./interfaces";
+import { makeWithPgClientFromPool, makeAddJob } from "./helpers";
+import { migrate } from "./migrate";
+import { defaultLogger } from "./logger";
+import { withReleasers, assertPool } from "./runner";
+
+const processPublisherOptions = async (options: WorkerUtilsOptions) => {
+  const { logger = defaultLogger } = options;
+  return withReleasers(async (releasers, _) => {
```
benjie: The second argument is the `release` method; it has to be called at some point but it's being discarded here. I think you might want to return it here so you can call it from the `end` method instead.
```suggestion
  return withReleasers(async (releasers, release) => {
```

src/worker_utils.ts
```diff
@@ -0,0 +1,62 @@
...
+    const pgPool: Pool = await assertPool(options, releasers, logger);
+
+    const withPgClient = makeWithPgClientFromPool(pgPool);
+
+    // Migrate
+    await withPgClient(client => migrate(client));
```
benjie: I think we should move this to a `migrate` method; calling this every time you call `require('graphile-worker').addJob(...)` could get expensive.

src/worker_utils.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    return { pgPool, withPgClient };
```
benjie:
```suggestion
    return { pgPool, withPgClient, release };
```

src/worker_utils.ts
```diff
@@ -0,0 +1,62 @@
...
+  });
+};
+
+export class WorkerUtils {
+  private dbTools: Promise<{
+    withPgClient: <T>(
+      callback: (pgClient: PoolClient) => Promise<T>
+    ) => Promise<T>;
+    pgPool: Pool;
```
benjie:
```suggestion
    pgPool: Pool;
    release: Release;
```

src/worker_utils.ts
```diff
@@ -0,0 +1,62 @@
...
+    this.dbTools = processPublisherOptions(options);
+  }
+
+  public async addJob(
+    identifier: string,
+    payload: any = {},
+    options: TaskOptions = {}
+  ): Promise<Job> {
+    const { withPgClient } = await this.dbTools;
+    const addJob = makeAddJob(withPgClient);
```
benjie: Let's go with a different implementation; the `addJob` function should enqueue promises (maybe via the `deferred` pattern?) and the constructor should trigger a "startup routine" which awaits for the dbTools, then overwrites the `addJob`/etc methods with the new methods, and then replays all the deferred actions. That way once startup is complete all calls to `addJob` will be directly to the underlying `addJob` function with no `await this.dbTools` overhead.

Though TBH this is a performance optimisation, the API is fine and that's what really matters initially.

src/worker_utils.ts
```diff
@@ -0,0 +1,62 @@
...
+    return addJob(identifier, payload, options);
+  }
+
+  async end(): Promise<void> {
+    const { pgPool } = await this.dbTools;
+    await pgPool.end();
```
benjie: This is not the right thing to do - we might not own the `pgPool`, so ending it could lead to very unexpected results for the user. This is where the `dbTools.release()` method would be called, which handles this complexity for you.

```suggestion
    (await this.dbTools).release();
```

benjie:
@mrmurphy If you allow me to make changes to your branch I can work towards getting this merged.

mrmurphy:
Yes! Thank you. I’m sorry, I’ve been swamped with work, and I’m pretty slow with typescript. This has been on my todo list, but slipping down repeatedly.

benjie:
@mrmurphy I still don't have permission to commit these changes to your branch.

mrmurphy:
Ah sorry, I didn't realize you were actually requesting GH permissions 😅. I'll get that changed.

mrmurphy:
@benjie Invited!

benjie:
I may have gone a little overboard with the refactoring... Took me a while to figure out where I wanted the API to be, but I think it's pretty nice now!

src/worker_utils.ts
benjie: Added a migrate method.

src/worker_utils.ts
benjie: I got rid of the class entirely in the end; turns out that WorkerUtils is really just a helpers object.

PR #63, opened 2020-01-07
feat: allow existing jobs to be updated and removed

### LINKED ISSUE #58: Cancel/modify delayed jobs
I'm exploring a case where I need to be able to somehow cancel or modify jobs in the graphile worker queue which have been delayed using the `run_at` argument. Looking at the source, it seems this would be trivial to achieve by deleting from `jobs` by id, but because the id is a serial, I would need to store it somewhere. If the id could be pre-generated (as a uuid4 or something similar) and passed to the `add_job` function, then I could generate the same id again in order to delete or update the job. So the changes would be 

- use a uuid for the jobs primary key instead of bigserial
- default it to `gen_random_uuid` but allow the id to be provided as an optional argument to `add_job`

Is there some reason I haven't considered, why this would be a bad idea? I'm assuming the insert performance would be better if the sequence was removed, but unsure how it would impact reads from the job table.

### LINKED ISSUE #59: Debouncing jobs
We're currently using https://github.com/timgit/pg-boss in production, and it has been solid, but the API is a bit complex. I love how simple the API for "worker" is, and I'm inclined to give it a try.

One feature we use regularly in Pg-boss that doesn't seem to be present yet in Worker is debouncing a job by name.

For example, I have a cron in my app that schedules a background job every 30 seconds, but if that job is already running, or if there's already one queued, I don't want to add another.

Would adding something like this seem inline with your goals for this project @benjie?

### LINKED ISSUE #61: RFC: Allow deletion and update of scheduled jobs
## Objective

- allow jobs scheduled for the future via `run_at` to be cancelled or updated prior to being run
- facilitate “singleton” jobs, where repeated scheduling of the same task results in only one eventual run
- Maintain full backwards compatibility

## Motivation

This change solves the issues raised in #58 and #59 - use cases:

### Update a scheduled job (#58)

- A “send push notification” task is scheduled for the start of an end user’s preferred window, e.g. 9am-5pm
- Before the notification is sent, the user changes their preferred window settings
- The previously scheduled notification should be updated to reflect the new preference

### Cancel a scheduled job (#58)

- A “send push notification” task is scheduled for a user as above
- Before the notification is sent, it is invalidated by a separate system
- The notification task should be deleted

### Debounce a cron-based job (#59)

- A cron process schedules a background job every 30 seconds
- If the job is already running, or if there's already one queued, an additional job should not be queued

#### `graphile_worker.add_job` 

An optional `job_key` argument will be added to this function, and it will be changed to use plpgsql to allow conditional logic. New function signature:

```diff
 create function graphile_worker.add_job(
   identifier text, 
   payload json = '{}', 
   queue_name text = public.gen_random_uuid()::text, 
   run_at timestamptz = now(), 
   max_attempts int = 25
+  ,job_key text = null
 ) returns graphile_worker.jobs;
```

##### If job_key is not provided

- The `add_job` behaviour will remain unchanged as a simple `insert` into `graphile_worker.jobs`

##### If job_key is provided

- The new job will overwrite **all details** of an existing job with the same key, using `insert ... on conflict`, only if the job is not locked for processing
- If the previous job is locked for processing, or completed, a new job will be scheduled
- ~~Note there is an edge case here, where a job has already been retrieved for processing, but not completed. In this case there is no way to schedule a new job, as would happen if the previous one was complete. The `on conflict ... where` clause will prevent the job details from changing, and the function will return an empty set - the calling code can use this return value to handle the case as needed. This case will be documented.~~ See https://github.com/graphile/worker/issues/61#issuecomment-568247838
- `jobs.attempts` will be reset to zero and `jobs.last_error` to null, under the assumption that past failures should not impact the lifecycle of the newly updated job.
- In a future update, a boolean argument could be added to determine whether existing jobs should be overwritten or not, i.e. on conflict `do update` vs `do nothing`

#### `graphile_worker.remove_job`

This new function will allow deletion of an existing job by `job_key`. Function signature:

```
create function remove_job(
  job_key text
) returns graphile_worker.jobs;
```

- Only jobs not already locked may be deleted. This means
  - jobs which a worker has already retrieved for processing cannot be deleted
  - there is no need to "unlock" the `job_queues` table the way [`complete_job`](https://github.com/graphile/worker/blob/17c420ca86f8609484eafec8fbb252eebbbc6730/sql/000001.sql#L135-L137) does
- If the `job_key` doesn't exist, or the job is already in progress, the function will do nothing and return an empty set

### Schema changes

1. Add a new nullable unique `key` column to the `graphile_worker.jobs` table:

   `alter table graphile_worker.jobs add column key text unique;`

   Full vs partial indexing for this column will be decided after profiling each option.

   Other column names considered were 

   - `idempotency_id` (misleading as job details may change if other arguments differ)
   - `job_id` (already used as a function argument, clashes with `jobs.id`)
   - `task_id` (clashes with `jobs.task_identifier`)

2. Add `locked_by` and `locked_at` columns to the jobs table, as per https://github.com/graphile/worker/issues/61#issuecomment-568247838

### Tests

New tests will prove that

- passing the same `job_key` to `add_job` multiple times results in one scheduled job
- future-scheduled jobs can be removed via `remove_job` and do not run
- jobs scheduled for now are not deleted or updated

### Performance implications

A write performance impact is expected from the additional column and unique index. However, this should only affect the opt-in new behaviour - if `job_key` is not specified, performance should stay the same as before.

Before/after performance characteristics will be determined and documented via the bundled [perfTest](https://github.com/graphile/worker/#performance) command.

### Alternatives considered

- Change the `jobs.id` to text, and allow the user to specify it, instead of a separate identifier (`job_key`). Rejected due to unknown impact on read performance, and backwards incompatibility
- Store `jobs.id` value as returned by `add_job` in a separate table, external to the library, and use that to update/remove jobs. Rejected due to additional complexity and statefulness required in the implementing code, and because the problem is general enough to be a valuable library addition

### Questions / discussion

- Is there any solution to the edge case noted when updating a job which is already processing?
- Which additional tests, if any, should be written?

gregplaysguitar:
See proposal and discussion in #61

### Performance impact

Running the existing performance test script shows no impact from these changes - running in Docker (see #64) and repeated 3 times for each case, the average latency was ~2ms each time and total time to run jobs ~30s.

### `jobs.key` index

There was some discussion over whether this should be implemented as a unique partial index, since it is likely to contain a lot of null values. This would mean we couldn't use an `on conflict` clause when inserting though, so I've used a regular unique constraint

### Checklist

- [x] Update `add_job` behaviour
- [x] Implement `remove_job` function
- [x] Determine whether the jobs.key unique index should be partial
- [x] Handle issues in Postgres 9.6 
- [x] Document performance impact
- [x] Update readme
- [x] pg functions strict, volatile etc  https://github.com/graphile/worker/pull/63#discussion_r365235334
- [x] implement "deferred" pattern in tests https://github.com/graphile/worker/pull/63#discussion_r365188493

---

Closes #58 
Closes #59 
Closes #61

### DIFF
--- a/__tests__/main.runTaskListOnce.test.ts
+++ b/__tests__/main.runTaskListOnce.test.ts
@@ -268,7 +268,7 @@ test("allows update of pending jobs", () =>
     );
     expect(updatedJobs).toHaveLength(1);
     const updatedJob = updatedJobs[0];
-    expect(updateJob.id).toEqual(job.id);
+    expect(updatedJob.id).toEqual(job.id);
     expect(updatedJob.run_at).toEqual(now);
 
     // Run the task
--- a/__tests__/main.runTaskListOnce.test.ts
+++ b/__tests__/main.runTaskListOnce.test.ts
@@ -342,10 +342,10 @@ test("schedules a new job if existing is being processed", () =>
     );
 
     // check there are now two jobs scheduled
-    const { rows: jobs } = await pgClient.query(
-      `select * from graphile_worker.jobs`
-    );
-    expect(jobs).toHaveLength(2);
+
+    expect(
+      (await pgClient.query(`select * from graphile_worker.jobs`)).rows
+    ).toHaveLength(2);
 
     // wait for the original job to complete - note this picks up the new job,
     // because the worker checks again for pending jobs at the end of each run
@@ -520,9 +520,70 @@ test("job details do not change unless specified in update", () =>
     });
   }));
 
-test("pending jobs can be removed", () => {});
+test("pending jobs can be removed", () =>
+  withPgClient(async pgClient => {
+    await reset(pgClient);
 
-test("jobs in progress can not be removed", () => {});
+    const tasks: TaskList = {
+      job1: jest.fn(() => {}),
+    };
+
+    // Schedule a job
+    await pgClient.query(
+      `select graphile_worker.add_job('job1', '{"a": "1"}', job_key := 'abc')`
+    );
+
+    // Assert that it has an entry in jobs / job_queues
+    expect(
+      (await pgClient.query(`select * from graphile_worker.jobs`)).rows
+    ).toHaveLength(1);
+
+    // remove the job
+    await pgClient.query(`select graphile_worker.remove_job('abc')`);
+
+    // check no jobs run
+    await runTaskListOnce(tasks, pgClient);
+    expect(tasks.job1).not.toHaveBeenCalled();
+  }));
+
+test("jobs in progress cannot be removed", () =>
+  withPgClient(async pgClient => {
+    await reset(pgClient);
+
+    const tasks: TaskList = {
+      job1: jest.fn(async () => {
+        await sleep(50);
+      }),
+    };
+
+    // Schedule a job
+    await pgClient.query(
+      `select graphile_worker.add_job('job1', '{"a": 123}', job_key := 'abc')`
+    );
+
+    // check it was inserted
+    expect(
+      (await pgClient.query(`select * from graphile_worker.jobs`)).rows
+    ).toHaveLength(1);
+
+    const promise = runTaskListOnce(tasks, pgClient);
+    // wait for it to be picked up for processing
+    await sleep(20);
+    expect(tasks.job1).toHaveBeenCalledTimes(1);
+
+    // attempt to remove the job
+    await pgClient.query(`select graphile_worker.remove_job('abc')`);
+
+    // check it was not removed
+    expect(
+      (await pgClient.query(`select * from graphile_worker.jobs`)).rows
+    ).toHaveLength(1);
+
+    // wait for the original job to complete
+    await promise;
+    expect(tasks.job1).toHaveBeenCalledTimes(1);
+    expect(tasks.job1).toHaveBeenCalledWith({ a: 123 }, expect.any(Object));
+  }));
 
 test("runs jobs asynchronously", () =>
   withPgClient(async pgClient => {
--- a/__tests__/main.runTaskListOnce.test.ts
+++ b/__tests__/main.runTaskListOnce.test.ts
@@ -335,7 +335,6 @@ test("schedules a new job if existing is being processed", () =>
     );
 
     // run the job
-    await sleep(20); // Give PostgreSQL connections a moment to synchronize
     const promise = runTaskListOnce(tasks, pgClient);
 
     // wait for it to be picked up for processing
--- a/__tests__/main.runTaskListOnce.test.ts
+++ b/__tests__/main.runTaskListOnce.test.ts
@@ -471,7 +471,7 @@ test("job details do not change unless specified in update", () =>
     expect(jobs).toHaveLength(1);
     expect(jobs[0]).toMatchObject(original);
 
-    // update job, but don't provided any new details
+    // update job, but don't provide any new details
     await pgClient.query(
       `select graphile_worker.add_job(
         'job1',
--- a/__tests__/main.runTaskListOnce.test.ts
+++ b/__tests__/main.runTaskListOnce.test.ts
@@ -433,7 +433,6 @@ test("schedules a new job if the existing is pending retry", () =>
   }));
 
 test("job details do not change unless specified in update", () =>
-  // queue_name, payload, task_identifier,
   withPgClient(async pgClient => {
     await reset(pgClient);
 
--- a/__tests__/main.runTaskListOnce.test.ts
+++ b/__tests__/main.runTaskListOnce.test.ts
@@ -1,23 +1,9 @@
 import { withPgClient, reset, sleepUntil, jobCount } from "./helpers";
 import { TaskList, Task, Worker } from "../src/interfaces";
 import { runTaskListOnce } from "../src/main";
-import deferred, { Deferred } from "../src/deferred";
+import defer, { Deferred } from "../src/deferred";
 
 const sleep = (ms: number) => new Promise(resolve => setTimeout(resolve, ms));
-
-interface Defer {
-  promise: Promise<void>;
-  resolve: () => void;
-}
-
-const defer = (): Defer => {
-  let resolve: Defer["resolve"];
-  const promise = new Promise<void>(_resolve => {
-    resolve = _resolve;
-  });
-  // @ts-ignore error TS2454: Variable 'resolve' is used before being assigned.
-  return { resolve, promise };
-};
 
 test("runs jobs", () =>
   withPgClient(async pgClient => {
@@ -333,13 +319,13 @@ test("schedules a new job if existing is being processed", () =>
   withPgClient(async pgClient => {
     await reset(pgClient);
 
-    const defers: Defer[] = [];
+    const defers: Deferred[] = [];
 
     const tasks: TaskList = {
       job1: jest.fn(async () => {
         const deferred = defer();
         defers.push(deferred);
-        return deferred.promise;
+        return deferred;
       }),
     };
 
@@ -625,9 +611,9 @@ test("runs jobs asynchronously", () =>
     await pgClient.query(`select graphile_worker.add_job('job1', '{"a": 1}')`);
 
     // Run the task
-    let jobPromise: Deferred<void> | null = null;
+    let jobPromise: Deferred | null = null;
     const job1: Task = jest.fn(() => {
-      jobPromise = deferred();
+      jobPromise = defer();
       return jobPromise;
     });
     const tasks: TaskList = {
@@ -693,9 +679,9 @@ test("runs jobs in parallel", () =>
     );
 
     // Run the task
-    const jobPromises: Array<Deferred<void>> = [];
+    const jobPromises: Array<Deferred> = [];
     const job1: Task = jest.fn(() => {
-      const jobPromise = deferred();
+      const jobPromise = defer();
       jobPromises.push(jobPromise);
       return jobPromise;
     });
@@ -772,9 +758,9 @@ test("single worker runs jobs in series, purges all before exit", () =>
     );
 
     // Run the task
-    const jobPromises: Array<Deferred<void>> = [];
+    const jobPromises: Array<Deferred> = [];
     const job1: Task = jest.fn(() => {
-      const jobPromise = deferred();
+      const jobPromise = defer();
       jobPromises.push(jobPromise);
       return jobPromise;
     });
@@ -820,9 +806,9 @@ test("jobs added to the same queue will be ran serially (even if multiple worker
     );
 
     // Run the task
-    const jobPromises: Array<Deferred<void>> = [];
+    const jobPromises: Array<Deferred> = [];
     const job1: Task = jest.fn(() => {
-      const jobPromise = deferred();
+      const jobPromise = defer();
       jobPromises.push(jobPromise);
       return jobPromise;
     });
--- a/__tests__/main.runTaskListOnce.test.ts
+++ b/__tests__/main.runTaskListOnce.test.ts
@@ -335,11 +335,11 @@ test("schedules a new job if existing is being processed", () =>
     );
 
     // run the job
-    await sleep(20); // Give PostgreSQL connections a moment to synchronise
+    await sleep(20); // Give PostgreSQL connections a moment to synchronize
     const promise = runTaskListOnce(tasks, pgClient);
 
     // wait for it to be picked up for processing
-    await sleep(20);
+    await sleepUntil(() => defers.length > 0);
     expect(tasks.job1).toHaveBeenCalledTimes(1);
 
     // attempt to update the job - it should schedule a new one instead
@@ -355,7 +355,7 @@ test("schedules a new job if existing is being processed", () =>
     // wait for the original job to complete - note this picks up the new job,
     // because the worker checks again for pending jobs at the end of each run
     defers[0].resolve(); // complete first job
-    await sleep(20); // wait for second job to be picked up
+    await sleepUntil(() => defers.length > 1); // wait for second job to be picked up
     defers[1].resolve(); // complete second job
     await promise; // wait for all jobs to finish
 
@@ -567,9 +567,12 @@ test("jobs in progress cannot be removed", () =>
   withPgClient(async pgClient => {
     await reset(pgClient);
 
+    let deferred: Deferred | null = null;
+
     const tasks: TaskList = {
       job1: jest.fn(async () => {
-        await sleep(50);
+        deferred = defer();
+        return deferred;
       }),
     };
 
@@ -585,7 +588,7 @@ test("jobs in progress cannot be removed", () =>
 
     const promise = runTaskListOnce(tasks, pgClient);
     // wait for it to be picked up for processing
-    await sleep(20);
+    await sleepUntil(() => !!deferred);
     expect(tasks.job1).toHaveBeenCalledTimes(1);
 
     // attempt to remove the job
@@ -597,7 +600,9 @@ test("jobs in progress cannot be removed", () =>
     ).toHaveLength(1);
 
     // wait for the original job to complete
+    deferred!.resolve();
     await promise;
+
     expect(tasks.job1).toHaveBeenCalledTimes(1);
     expect(tasks.job1).toHaveBeenCalledWith({ a: 123 }, expect.any(Object));
   }));
@@ -827,7 +832,7 @@ test("jobs added to the same queue will be ran serially (even if multiple worker
 
     for (let i = 0; i < 5; i++) {
       // Give the other workers a chance to interfere (they shouldn't)
-      sleep(50);
+      await sleep(50);
 
       await sleepUntil(() => jobPromises.length >= i + 1);
       expect(jobPromises).toHaveLength(i + 1);
--- a/__tests__/main.runTaskListOnce.test.ts
+++ b/__tests__/main.runTaskListOnce.test.ts
@@ -268,6 +268,7 @@ test("allows update of pending jobs", () =>
     );
     expect(updatedJobs).toHaveLength(1);
     const updatedJob = updatedJobs[0];
+    expect(updateJob.id).toEqual(job.id);
     expect(updatedJob.run_at).toEqual(now);
 
     // Run the task
--- a/__tests__/main.runTaskListOnce.test.ts
+++ b/__tests__/main.runTaskListOnce.test.ts
@@ -238,7 +238,7 @@ test("allows update of pending jobs", () =>
 
     // Schedule a future job - note incorrect payload
     const runAt = new Date();
-    runAt.setSeconds(runAt.getSeconds() + 3);
+    runAt.setSeconds(runAt.getSeconds() + 60);
 
     await pgClient.query(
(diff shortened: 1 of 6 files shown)


sql/000002.sql
```diff
@@ -0,0 +1,165 @@
...
+create or replace function graphile_worker.fail_job(worker_id text, job_id bigint, error_message text) returns graphile_worker.jobs as $$
+declare
+  v_row graphile_worker.jobs;
+begin
+  update graphile_worker.jobs
+    set
+      last_error = error_message,
+      run_at = greatest(now(), run_at) + (exp(least(attempts, 10))::text || ' seconds')::interval,
+      locked_by = null,
+      locked_at = null
```
gregplaysguitar: Added these lines

sql/000002.sql
```diff
@@ -0,0 +1,165 @@
...
+    set
+      locked_by = worker_id,
+      locked_at = v_now
+    where job_queues.queue_name = v_queue_name;
+
+  update graphile_worker.jobs
+    set
+      attempts = attempts + 1,
+      locked_by = worker_id,
+      locked_at = v_now
```
gregplaysguitar: Added these lines

__tests__/main.runTaskListOnce.test.ts
```diff
@@ -34,15 +34,15 @@ test("runs jobs", () =>
...
     const job1: Task = jest.fn(o => {
       expect(o).toMatchInlineSnapshot(`
-Object {
-  "a": 1,
-}
-`);
+        Object {
+          "a": 1,
+        }
+      `);
```
gregplaysguitar: Prettier formatting noise

sql/000002.sql
```diff
@@ -0,0 +1,165 @@
...
+
+-- update add_job behaviour to meet new requirements
+drop function graphile_worker.add_job;
+create function graphile_worker.add_job(
+  identifier text,
+  payload json = null,
+  queue_name text = null,
+  run_at timestamptz = null,
+  max_attempts int = null,
+  job_key text = null
```
gregplaysguitar: All optional arguments default to null because we need to be able to keep their existing values in the db, when doing an on conflict update - see coalesce calls below

sql/000002.sql
```diff
@@ -0,0 +1,165 @@
...
+    insert into graphile_worker.jobs(task_identifier, payload, queue_name, run_at, max_attempts)
+      values(
+        identifier,
+        coalesce(payload, '{}'),
+        coalesce(queue_name, public.gen_random_uuid()::text),
+        coalesce(run_at, now()),
+        coalesce(max_attempts, 25)
+      )
+      returning *
+      into v_job;
```
gregplaysguitar: Identical to old previous behaviour

sql/000002.sql
```diff
@@ -0,0 +1,165 @@
...
+          identifier,
+          coalesce(payload, '{}'),
+          coalesce(queue_name, public.gen_random_uuid()::text),
+          coalesce(run_at, now()),
+          coalesce(max_attempts, 25),
+          job_key
+        )
+        returning *
+        into v_job;
+    end if;
```
gregplaysguitar: Rationale here https://github.com/graphile/worker/issues/61#issuecomment-568247838

__tests__/main.runTaskListOnce.test.ts
```diff
@@ -225,6 +225,366 @@ test("supports future-scheduled jobs", () =>
...
+    );
+    expect(tasks.job1).toHaveBeenNthCalledWith(
+      2,
+      { succeed: true },
+      expect.any(Object)
+    );
+  }));
+
+test("job details do not change unless specified in update", () =>
+  // queue_name, payload, task_identifier,
```
gregplaysguitar:
```suggestion
```

sql/000002.sql
```diff
@@ -0,0 +1,183 @@
...
+        values(
+          identifier,
+          coalesce(payload, '{}'),
+          coalesce(queue_name, public.gen_random_uuid()::text),
+          coalesce(run_at, now()),
+          coalesce(max_attempts, 25),
+          job_key
+        )
+        returning *
+        into v_job;
```
benjie: Since this is the exact code as at the start, for ease of maintenance / DRY can we flip the logic? Something like

```sql
if job_key is not null then
  insert ... on conflict ... returning * into v_job;
  if not (v_job is null) then -- IMPORTANT: this is not the same as `if v_job is not null`
    return v_job;
  else
    update jobs set key = null...;
    -- We'll re-attempt insert without ON CONFLICT below
  end if;
end if;

insert into ... values ... returning * into v_job;
return v_job;
```

sql/000002.sql
```diff
@@ -0,0 +1,183 @@
+-- TODO evaluate perf re partial indexing instead of full unique
+alter table graphile_worker.jobs add column key text unique;
+
+alter table graphile_worker.jobs add locked_at timestamptz;
+alter table graphile_worker.jobs add locked_by text;
```
benjie: One of the potential advantages this may grant us is the ability to set a concurrency for a particular queue. Currently all queues have concurrency 1. Before merging I'd like to think about this a little deeper (your thoughts are welcome too) because designing the schema in such a way as to easily facilitate this later would be better than schema churn.

sql/000002.sql
```diff
@@ -0,0 +1,183 @@
...
+
+--- implement new remove_job function
+
+create function graphile_worker.remove_job(
+  job_key text
+) returns graphile_worker.jobs as $$
+  delete from graphile_worker.jobs
+    where key = job_key
+    and locked_at is null
+  returning *;
```
benjie: I originally thought this should `returns void` to make it idempotent; but returning the deleted job leaves more possibilities open so let's stick with this :+1:

sql/000002.sql
```diff
@@ -0,0 +1,183 @@
...
+$$ language sql;
```
benjie: In this case if `job_key` is null then the result is null so we can skip the function call (`STRICT`) and it's volatile which I prefer to be explicit about (since it's the default, I prefer that we explicitly state we decided it SHOULD be volatile, rather than that it's just volatile because we weren't paying close enough attention).

```suggestion
$$ language sql strict volatile;
```

Could you make similar edits to the other functions if necessary please? Very few will be `strict`, mind.

sql/000002.sql
benjie: We should make sure the existing jobs table is updated to honour this, in case there are any jobs in flight. I've not sufficiently validated this SQL, so please check it carefully!

```suggestion
alter table graphile_worker.jobs add locked_by text;

update graphile_worker.jobs
  set locked_at = q.locked_at, locked_by = q.locked_by
  from graphile_worker.job_queues q
  where q.queue_name = jobs.queue_name
  and q.locked_at is not null;
```

sql/000002.sql
```diff
@@ -0,0 +1,183 @@
...
+
+
+-- Update other functions to handle locked_at denormalisation
+
+create or replace function graphile_worker.get_job(worker_id text, task_identifiers text[] = null, job_expiry interval = interval '4 hours') returns graphile_worker.jobs as $$
+declare
+  v_job_id bigint;
+  v_queue_name text;
+  v_default_job_max_attempts text = '25';
```
benjie:
```suggestion
```

What was Past Benjie doing?!

sql/000002.sql
```diff
@@ -0,0 +1,183 @@
...
+  v_row graphile_worker.jobs;
+  v_now timestamptz = now();
```
benjie: Technically `now()` will never change within a transaction, so this is unnecessary. I think it makes things more explicit though, so we'll keep it.

sql/000002.sql
```diff
@@ -0,0 +1,183 @@
...
+  select job_queues.queue_name, jobs.id into v_queue_name, v_job_id
+    from graphile_worker.jobs
+    inner join graphile_worker.job_queues using (queue_name)
+    where (job_queues.locked_at is null or job_queues.locked_at < (v_now - job_expiry))
+    and run_at <= v_now
+    and attempts < max_attempts
+    and (task_identifiers is null or task_identifier = any(task_identifiers))
+    order by priority asc, run_at asc, id asc
+    limit 1
+    for update of job_queues
```
benjie: We're now expecting to update both tables, so we should hint this:

```suggestion
    for update of job_queues, jobs
```

sql/000002.sql
```diff
@@ -0,0 +1,183 @@
...
+      locked_at = v_now
+    where id = v_job_id
+    returning * into v_row;
+
+  return v_row;
+end;
+$$ language plpgsql;
+
+-- I was unsuccessful, re-schedule the job please
+create or replace function graphile_worker.fail_job(worker_id text, job_id bigint, error_message text) returns graphile_worker.jobs as $$
```
benjie: Note to self: `complete_job` doesn't need to change since it deletes the row.

__tests__/main.runTaskListOnce.test.ts
```diff
@@ -225,6 +225,365 @@ test("supports future-scheduled jobs", () =>
...
+    expect(updatedJobs).toHaveLength(1);
+    const updatedJob = updatedJobs[0];
+    expect(updatedJob.run_at).toEqual(now);
+
+    // Run the task
+    await runTaskListOnce(tasks, pgClient);
+    expect(tasks.job1).toHaveBeenCalledTimes(1);
+  }));
+
+test("schedules a new job if existing is completed", () =>
```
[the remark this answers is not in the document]
gregplaysguitar: I would say that the actual deletion from the jobs table is an implementation detail, this test proves the behaviour from the perspective of the public API, so I think it has value.

sql/000002.sql
```diff
@@ -0,0 +1,183 @@
+-- TODO evaluate perf re partial indexing instead of full unique
+alter table graphile_worker.jobs add column key text unique;
```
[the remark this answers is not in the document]
gregplaysguitar: I think this is a good idea, if a user passed an empty string they might expect it to behave the same as passing null, so adding a check avoids any confusion. I assume the perf impact is negligible

sql/000002.sql
```diff
@@ -0,0 +1,183 @@
...
+        -- update run_at if argument is provided. If not, and there has been an
+        -- error, assume run_at reflects the error retry backoff, so reset it.
+        -- If no errors, maintain existing value by default
+        run_at=coalesce(
+          add_job.run_at,
+          (case when jobs.attempts > 0
+            then now()
+            else jobs.run_at
+          end)
+        ),
```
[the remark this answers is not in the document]
gregplaysguitar: Yup, makes sense. I'll update

__tests__/main.runTaskListOnce.test.ts
```diff
@@ -225,6 +225,365 @@ test("supports future-scheduled jobs", () =>
...
+    expect(
+      (await pgClient.query(`select * from graphile_worker.jobs`)).rows
+    ).toHaveLength(1);
+
+    // remove the job
+    await pgClient.query(`select graphile_worker.remove_job('abc')`);
+
+    // check no jobs run
+    await runTaskListOnce(tasks, pgClient);
+    expect(tasks.job1).not.toHaveBeenCalled();
```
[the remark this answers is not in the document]
gregplaysguitar: Hmm, I think it's better to test a public API wherever possible, in this case being whether the job was actually run or not. We can make it typo-proof by defining the job name as a variable at the top of the test? Happy to add the jobs table check too if you want, but personally I think what I've got here is the higher-value test

sql/000002.sql
gregplaysguitar: This makes sense, but we were already updating both tables - https://github.com/graphile/worker/blob/master/sql/000001.sql#L117-L120

Is there something particular about updating the locked_* columns? I'm just keen to understand what's going on here as I'm not that familiar with `skip locked` etc

sql/000002.sql
gregplaysguitar: Could you elaborate on this? Which part is unnecessary

sql/000002.sql
benjie: This article has a great explanation: https://www.2ndquadrant.com/en/blog/what-is-select-skip-locked-for-in-postgresql-9-5/

benjie: You're right; we don't need to add `jobs` to the locked list here - it's an unnecessary performance overhead. Locking the `job_queue` itself is effectively a sufficient mutex that we're allowed to update the job and there'll be no concurrent update on it.

__tests__/main.runTaskListOnce.test.ts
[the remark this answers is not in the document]
benjie: We're testing the public API with the `remove_job` call; the check should be that the public API did what it said and deleted the job (i.e. no rows). This current check just tests that it doesn't run on the next `runTaskListOnce` which could be achieved by setting attempts=max_attempts or bumping run_at to somewhere in the future or locking the queue (maybe?) or various other approaches, none of which remove the job.

sql/000002.sql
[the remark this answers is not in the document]
benjie: Yeah; agreed - lets add it. It's a tiny cost on write only (no read cost), so I think it's worth it :+1:

sql/000002.sql
benjie: You refactored the code to use the `v_now` variable rather than using the `now()` call directly in various different locations within the function (you can look [at the old code](https://github.com/graphile/worker/blob/52b776c1d83ffdbbd29cf053c80bfcabe29b0fb3/sql/000001.sql#L84-L124) to see us calling `now()` in a couple different statements in the function without a `v_now` variable).

In case you're not familiar, here's where `NOW()` is documented:

https://www.postgresql.org/docs/current/functions-datetime.html#FUNCTIONS-DATETIME-CURRENT

Of relevance (emphasis mine):

> These SQL-standard functions all return values based on **the start time of the current transaction**:

and

> Since these functions return the start time of the current transaction, **their values do not change during the transaction**. 

Functions always run within a single transaction, so all `now()` calls in the function will have the same value and there's no need to factor them out to a `v_now` variable. That said, I don't think this does any harm and it makes the code clearer for people not familiar with this concept so I'm inclined to keep it.

sql/000002.sql
```diff
@@ -0,0 +1,176 @@
...
+        payload,
+        coalesce(queue_name, public.gen_random_uuid()::text),
+        run_at,
+        max_attempts,
+        job_key
+      )
+      on conflict (key) do update set
+        -- update all job details other than queue_name, which we want to keep
+        -- the same unless explicitly provided
+        task_identifier=identifier,
```
benjie: For this and the others, please reference the [`excluded` record](https://www.postgresql.org/docs/current/sql-insert.html#SQL-ON-CONFLICT), rather than the arguments directly (which can in some cases share the name with the columns and make it less clear which values are being used).

```suggestion
        task_identifier = excluded.task_identifier,
```

sql/000002.sql
```diff
@@ -0,0 +1,176 @@
...
+        payload=payload,
+        queue_name=coalesce(add_job.queue_name, jobs.queue_name),
```
benjie: Very sensible - don't change the queue_name unless it was requested (in case we're partitioning on it) :+1:

sql/000002.sql
gregplaysguitar: Ah right, I see. Thanks for the explanation :)

__tests__/main.runTaskListOnce.test.ts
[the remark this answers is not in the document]
gregplaysguitar: >  could be achieved by setting attempts=max_attempts or bumping run_at 

Good point

I'll defer to you on this then, are you happy to have both checks? Or do you want the no-run check removed?

sql/000002.sql
gregplaysguitar: Have done - I've made `fail_job` strict too, seeing as none of its arguments have defaults. The other functions have optional arguments which default to null.

__tests__/main.runTaskListOnce.test.ts
```diff
@@ -225,6 +225,365 @@ test("supports future-scheduled jobs", () =>
...
+    );
+  }));
+
+test("schedules a new job if existing is being processed", () =>
+  withPgClient(async pgClient => {
+    await reset(pgClient);
+
+    const tasks: TaskList = {
+      job1: jest.fn(async () => {
+        await sleep(50);
```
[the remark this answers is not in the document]
gregplaysguitar: Ok, I think I have this figured out. See https://github.com/graphile/worker/pull/63/commits/a970f16927771633a6f3f0ca71407ecbe464e244

I did have to add an extra sleep in the test body, because of the way the runner picks up a new job after the current one completes. 

Also, I can't see how to avoid a ts-ignore there - let me know if you have any ideas.

gregplaysguitar:
@benjie this is now ready for another review. Looking forward to your feedback!

benjie:
I've rebased on master; I had to drop the "temp docker merge" commit, but that was all. The diff versus the previous branch is:

```diff
diff --git c/README.md w/README.md
index 2666467..0a13ed1 100644
--- c/README.md
+++ w/README.md
@@ -440,6 +440,19 @@ You can skip parameters you don't need by using PostgreSQL's named parameter sup
 SELECT graphile_worker.add_job('reminder', run_at := NOW() + INTERVAL '2 days');
 ```
 
+**TIP**: if you want to run a job after a variable number of seconds, you can use
+interval multiplication; see `run_at` in this example:
+
+```sql
+SELECT graphile_worker.add_job(
+  $1, 
+  payload := $2,
+  queue_name := $3,
+  max_attempts := $4,
+  run_at := NOW() + ($5 * INTERVAL '1 second')
+);
+```
+
 **NOTE:** `graphile_worker.add_job(...)` requires database owner privileges
 to execute. To allow lower-privileged users to call it, wrap it inside a
 PostgreSQL function marked as `SECURITY DEFINER` so that it will run
diff --git c/perfTest/recreateDb.js w/perfTest/recreateDb.js
index 3a851f9..867e447 100644
--- c/perfTest/recreateDb.js
+++ w/perfTest/recreateDb.js
@@ -23,7 +23,9 @@ async function main() {
   }
 }
 
-main().catch(e => {
-  console.error(e);
-  process.exit(1);
-});
+main()
+  .then(() => pgPool.end())
+  .catch(e => {
+    console.error(e);
+    process.exit(1);
+  });
diff --git c/src/runner.ts w/src/runner.ts
index 6c8b240..70b54d8 100644
--- c/src/runner.ts
+++ w/src/runner.ts
@@ -139,11 +139,11 @@ export const run = async (options: RunnerOptions): Promise<Runner> => {
   return {
     async stop() {
       if (running) {
-        throw new Error("Runner is already stopped");
-      } else {
         running = false;
         await workerPool.release();
         await release();
+      } else {
+        throw new Error("Runner is already stopped");
       }
     },
     addJob: makeAddJob(withPgClient),
```

This is all expected changes from master :+1:

benjie:
This is now released in `graphile-worker@0.3.0-rc.0` AKA `graphile-worker@rc`. Once I've tested it in a few places I'll promote it to 0.3.0 (but probably not today).

gregplaysguitar:
Thanks, excited to start working with this in our app!

benjie:
Thanks for all your work and persisting through multiple review rounds!

PR #64, opened 2020-01-07
Add docker-compose config for local development

gregplaysguitar:
I've used this configuration while working on #63 but thought it was best as a separate PR. Basically this should make it easier to get up and running for anyone who already has docker - in my case I couldn't run the tests very easily on my local machine because I didn't have postgres installed.

In theory everything should still work as before, but I haven't been able to test that thoroughly


Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+COPY package.json yarn.lock ./
+RUN yarn install --frozen-lockfile && yarn cache clean
+
+COPY . .
+
+RUN yarn prepack
+
+ENTRYPOINT ["yarn"]
+CMD ["watch"]
```
gregplaysguitar: Since this is for local dev only, the default command is just to watch and rebuild on change

perfTest/init.js
```diff
@@ -0,0 +1,13 @@
+const { Pool } = require("pg");
```
gregplaysguitar: Moved this into a js file so that it could run on a remote db, without requiring psql

perfTest/run.sh
```diff
@@ -1,24 +1,30 @@
 #!/usr/bin/env bash
```
gregplaysguitar: This script should still run locally as before, although I haven't been able to test that thoroughly.

Dockerfile
```diff
@@ -0,0 +1,14 @@
+# common base image for development and production
+FROM node:10.18.0-alpine AS base
```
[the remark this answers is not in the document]
gregplaysguitar: Yeah good point, I'll change it. I just copied this line straight out of our app, which is pinned but gets updated via renovate, there's no reason I know of to pin versions for the usecase here

perfTest/run.sh
```diff
@@ -1,24 +1,30 @@
...
-# Reset the database
-dropdb graphile_worker_perftest || true;
-createdb graphile_worker_perftest;
+export DATABASE_URL=$TEST_CONNECTION_STRING || "graphile_worker_perftest"
+
+if [ -x "$(command -v createdb)" ]; then
+    # Reset the database if running locally
+    dropdb graphile_worker_perftest || true
+    createdb graphile_worker_perftest
+fi
```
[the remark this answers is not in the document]
gregplaysguitar: Ok, I've refactored this so it doesn't need dropdb/createdb, seems better to be consistent. Let me know if you see any issues with the new approach

Dockerfile
```diff
@@ -1,12 +1,11 @@
...
-FROM node:10.18.0-alpine AS base
+FROM node:12-alpine
 RUN apk add --no-cache bash
 WORKDIR /app
 
 COPY package.json yarn.lock ./
 RUN yarn install --frozen-lockfile && yarn cache clean
 
-COPY . .
+COPY ./__tests__ ./sql ./src ./perfTest ./tsconfig.json ./
```
benjie: Does this do the right thing? Docker `COPY` notes:

> If <src> is a directory, the entire contents of the directory are copied, including filesystem metadata.
> Note: The directory itself is not copied, just its contents.

but also:

> If multiple <src> resources are specified, either directly or due to the use of a wildcard, then <dest> must be a directory, and it must end with a slash /.

https://docs.docker.com/engine/reference/builder/#copy

Initially I thought this would be equivalent to:

```
cp ./__tests__/* ./sql/* ./src/* ./perfTest/* ./tsconfig.json DESTINATION
```

i.e. everything gets smooshed into one folder? But does this only happen when there's exactly one `src` argument (which is a directory), and if there's multiple src arguments then the behaviour differs? (This seems highly confusing!)

perfTest/recreateDb.js
```diff
@@ -0,0 +1,14 @@
+const { Pool } = require("pg");
+const { parse } = require("pg-connection-string");
+
+const config = parse(process.env.PERF_DATABASE_URL);
+// don't connect to the provided db, or we can't drop it
+const pgPool = new Pool({ ...config, database: undefined });
```
benjie: I think this won't work if a default database isn't available. (e.g. my username is "benjie" and there's no "benjie" database, so I think this would error for me).

I get around this by using a database name that I know almost everyone has: `template1`:

```suggestion
const pgPool = new Pool({ ...config, database: "template1" });
```

perfTest/recreateDb.js
```diff
@@ -0,0 +1,14 @@
...
+
+async function main() {
+  console.log(`Recreating database ${config.database}`);
+  await pgPool.query(`drop database if exists ${config.database};`);
+  await pgPool.query(`create database ${config.database};`);
```
benjie: Please escape these values, and do this with a single database connection to avoid race conditions. I think this might work:

```js
const pgClient = await pgPool.connect();
try {
  await pgClient.query(`drop database if exists ${pgClient.escape(config.database)};`);
  await pgClient.query(`create database ${pgClient.escape(config.database)};`);
} finally {
  pgClient.release();
}
```

perfTest/recreateDb.js
```diff
@@ -0,0 +1,14 @@
...
+}
+
+main();
```
benjie: Make sure we don't get unhandled promise rejection errors.

```suggestion
main().catch(e => {
  console.error(e);
  process.exit(1);
});
```

perfTest/recreateDb.js
```diff
@@ -0,0 +1,14 @@
+const { Pool } = require("pg");
+const { parse } = require("pg-connection-string");
+
+const config = parse(process.env.PERF_DATABASE_URL);
```
benjie:
```suggestion
if (!process.env.PERF_DATABASE_URL) {
  throw new Error("No PERF_DATABASE_URL setting detected; please don't call this script directly!");
}
const config = parse(process.env.PERF_DATABASE_URL);
```

perfTest/run.sh
```diff
@@ -1,24 +1,28 @@
...
-createdb graphile_worker_perftest;
+# if connection string not provided, assume postgres is available locally
+export PERF_DATABASE_URL=${TEST_CONNECTION_STRING-graphile_worker_perftest}
+
+# drop and recreate the test database
+node ./recreateDb.js
 
 # Install the schema
-DATABASE_URL="graphile_worker_perftest" node ../dist/cli.js --once
+DATABASE_URL=$PERF_DATABASE_URL node ../dist/cli.js --once
```
benjie: Should always use quotes to avoid shell expansion

```suggestion
DATABASE_URL="$PERF_DATABASE_URL" node ../dist/cli.js --once
```

perfTest/run.sh
```diff
@@ -1,24 +1,28 @@
...
 
 # How long does it take to start up and shut down?
-DATABASE_URL="graphile_worker_perftest" time node ../dist/cli.js --once
+DATABASE_URL=$PERF_DATABASE_URL time node ../dist/cli.js --once
```
benjie:
```suggestion
DATABASE_URL="$PERF_DATABASE_URL" time node ../dist/cli.js --once
```

perfTest/run.sh
```diff
@@ -1,24 +1,28 @@
...
 
 # Schedule the jobs
-psql -f init.sql graphile_worker_perftest;
+node ./init.js
 
 # Finally time the job execution
-DATABASE_URL="graphile_worker_perftest" time node ../dist/cli.js --once
+DATABASE_URL=$PERF_DATABASE_URL time node ../dist/cli.js --once
```
benjie:
```suggestion
DATABASE_URL="$PERF_DATABASE_URL" time node ../dist/cli.js --once
```

Dockerfile
gregplaysguitar: Hmm, good catch, I missed that somehow. FWIW I think just using `COPY . .` is better, it saves having to duplicate the paths, and given this is an image for local dev only I don't see it being an issue. But I'll separate out into separate `COPY` statements

perfTest/recreateDb.js
gregplaysguitar: That doesn't work, and I'm not really sure what the point of this is. What's the risk of not escaping, given we're already running `drop database` here? If you do want it escaped, can you let me know how I should go about it? I'm not sure how to do it sorry.

perfTest/recreateDb.js
gregplaysguitar: With this change, I get an error 

> error: source database "template1" is being accessed by other users

With undefined I don't get this. I tried changing the user name to one without a corresponding db though, and got the error as you expected. Any ideas?

perfTest/recreateDb.js
gregplaysguitar: See https://github.com/brianc/node-postgres/issues/33#issuecomment-1306277

perfTest/recreateDb.js
gregplaysguitar: Could we use `postgres` as the default? From the [docs:](https://www.postgresql.org/docs/current/creating-cluster.html) 

> The database server itself does not require the postgres database to exist, but many external utility programs assume it exists.

gregplaysguitar:
@benjie this is ready for review again.

perfTest/recreateDb.js
benjie: Sorry, turns out it's called `escapeIdentifier` rather than simply `escape`:

https://github.com/brianc/node-postgres/blob/3f6760c62ee2a901d374b5e50c2f025b7d550315/packages/pg/lib/client.js#L408-L411

The point of escaping is to avoid potential SQL injection - dropping a different database isn't the only thing you could do - your attack could be much more subtle and nefarious :smiling_imp:

docker-compose.yml
```diff
@@ -0,0 +1,23 @@
...
+
+services:
+  db:
+    image: postgres:11.5-alpine
+    tmpfs: /var/lib/postgresql/data
+    ports:
+      - "${PORT_DB-5432}:5432"
+    environment:
+      POSTGRES_DB: graphile_worker_test
+    container_name: graphile_worker_db
```
benjie: If this just used the `db` container_name would it conflict with other things on the system, or is the container_name local to this docker-compose somehow? (Docker noob here.)

perfTest/recreateDb.js
```diff
@@ -0,0 +1,28 @@
...
+const { parse } = require("pg-connection-string");
+
+if (!process.env.PERF_DATABASE_URL) {
+  throw new Error(
+    "No PERF_DATABASE_URL setting detected; please don't call this script directly!"
+  );
+}
+const config = parse(process.env.PERF_DATABASE_URL);
+// don't connect to the provided db, or we can't drop it
+const pgPool = new Pool({ ...config, database: "postgres" });
```
benjie: Not everyone has a `postgres` database (I learned this the hard way), so I think using `template1` is more compatible:

```suggestion
const pgPool = new Pool({ ...config, database: "template1" });
```

perfTest/recreateDb.js
```diff
@@ -0,0 +1,28 @@
...
+
+async function main() {
+  console.log(`Recreating database ${config.database}`);
+
+  const pgClient = await pgPool.connect();
+  try {
+    await pgPool.query(`drop database if exists ${config.database};`);
```
benjie:
```suggestion
    await pgClient.query(`drop database if exists ${pgClient.escapeIdentifier(config.database)};`);
```

perfTest/recreateDb.js
```diff
@@ -0,0 +1,28 @@
...
+    await pgPool.query(`create database ${config.database};`);
```
benjie:
```suggestion
    await pgClient.query(`create database ${pgClient.escapeIdentifier(config.database)};`);
```

perfTest/recreateDb.js
benjie: `pgClient` -> avoids a very unlikely but possible race condition due to pgPool potentially issuing two different clients which could cause synchronisation problems on a very overwhelmed database server.

`pgClient.escapeIdentifier` -> avoid SQL injection. Dropping a database based on envvars is already pretty bad, you might figure, but it's always worth limiting what someone can do to only what is expected.

perfTest/recreateDb.js
gregplaysguitar: Yeah, but when I tried this it doesn’t work in docker - I’m not sure why. See https://github.com/graphile/worker/pull/64#discussion_r367818939

I guess we might need to use template1 only if not running in docker? Feels a bit messy but can’t think of another solution right now. Let me know if you have any ideas

docker-compose.yml
gregplaysguitar: It needs to be globally unique if we want to execute commands in the container using docker (bypassing compose - see https://github.com/gregplaysguitar/worker/blob/55a7926731b3277fc57ce62e4c9b82109825d7ef/README.md#L618). 

The reason for that was that I was getting "the input device is not a TTY" when trying to run it via compose... however I just worked out that passing `-T` fixes that so I'll update. Definitely better to keep it local.

perfTest/recreateDb.js
gregplaysguitar: Actually, the error I was seeing has gone away now that I'm using a single client as suggested. So all good

perfTest/recreateDb.js
gregplaysguitar: Done - I do know about sql injection 🙂, just wasn't sure how it would work given the value is parsed from a url. `escapeIdentifier` makes perfect sense though, thanks.

perfTest/recreateDb.js
benjie: Hmmm... From what I've experienced both `undefined` and `"postgres"` cannot be relied on. `undefined` will definitely fail on my machine, I very deliberately do not have a default database because I run so many tools it ensures that I remember to connect to the right DB. `template1` is what we use in all the other tools because it's almost guaranteed to exist. Any idea what you have that is connected to the template1 DB and causing this issue? You shouldn't generally have anything connected to it. Try `select * from pg_stat_activity` (https://www.postgresql.org/docs/current/monitoring-stats.html#PG-STAT-ACTIVITY-VIEW) to see what it is.

If you've not factored in the `pgClient` changes suggested below (e.g. you're still using `pgPool.query` in `recreateDb.js`) then this might be the reason - the "race conditions" I mentioned (two clients connected to the same database at the same time, rather than using just one client).

perfTest/recreateDb.js
benjie: Ah; ignore my reply above - this is exactly what I figured was going on :+1:

docker-compose.yml
```diff
@@ -0,0 +1,21 @@
...
+      POSTGRES_DB: graphile_worker_test
+
+  app:
+    build:
+      context: .
+    volumes:
+      - ./__tests__:/app/__tests__
+      - ./sql:/app/sql
+      - ./src:/app/src
+      - ./perfTest:/app/perfTest
```
benjie: Why are we copying these files in the Dockerfile if we're also mounting them here?

perfTest/recreateDb.js
```diff
@@ -0,0 +1,29 @@
...
+  console.log(`Recreating database ${config.database}`);
+
+  try {
+    await pgClient.query(`create database ${dbName};`);
+  } finally {
+    pgClient.release();
+  }
+}
+
+main().catch(e => {
```
benjie:
```suggestion
main()
  .then(() => pgPool.end())
  .catch(e => {
```

benjie: Without this, I think the `recreateDb` script would hang?

perfTest/recreateDb.js
benjie: (You figured this out below.)

docker-compose.yml
benjie: :+1: Thanks for explanation

benjie:
All previous threads marked as resolved :+1:

docker-compose.yml
gregplaysguitar: They're copied so that the Dockerfile is usable without the docker-compose config - there's no reason you can't just run it directly with docker. But we also mount them here so that for local dev you don't have to rebuild the docker image every time you make a change. I can see how it's a bit redundant, but I think it'd be a bit odd to ship a Dockerfile that didn't work without specific directories mounted.

benjie: Got it; thanks :+1:

Issue #68, opened 2020-01-28
Unable to run perfTest with docker-compose


beepsoft:
I'm trying to run the perfTest according to the README.

I'm running it on macOS with Docker 19.03.4, Docker Compose 1.25.2

I ran
```
docker-compose up
```

Then in another terminal:
```
docker-compose exec app ./perfTest/run.sh
```

I always get this error:
```
> docker-compose exec app ./perfTest/run.sh
Recreating database graphile_worker_test
error: database "graphile_worker_test" already exists
    at Connection.parseE (/app/node_modules/pg/lib/connection.js:604:11)
    at Connection.parseMessage (/app/node_modules/pg/lib/connection.js:401:19)
    at Socket.<anonymous> (/app/node_modules/pg/lib/connection.js:121:22)
    at Socket.emit (events.js:223:5)
    at addChunk (_stream_readable.js:309:12)
    at readableAddChunk (_stream_readable.js:290:11)
    at Socket.Readable.push (_stream_readable.js:224:10)
    at TCP.onStreamRead (internal/stream_base_commons.js:181:23) {
  name: 'error',
  length: 103,
  severity: 'ERROR',
  code: '42P04',
  detail: undefined,
  hint: undefined,
  position: undefined,
  internalPosition: undefined,
  internalQuery: undefined,
  where: undefined,
  schema: undefined,
  table: undefined,
  column: undefined,
  dataType: undefined,
  constraint: undefined,
  file: 'dbcommands.c',
  line: '480',
  routine: 'createdb'
}
```

This is the output from `docker-compose up`:

```
Starting graphile-worker_db_1  ... done
Starting graphile-worker_app_1 ... done
Attaching to graphile-worker_db_1, graphile-worker_app_1
db_1   | The files belonging to this database system will be owned by user "postgres".
db_1   | This user must also own the server process.
db_1   | 
db_1   | The database cluster will be initialized with locale "en_US.utf8".
db_1   | The default database encoding has accordingly been set to "UTF8".
db_1   | The default text search configuration will be set to "english".
db_1   | 
db_1   | Data page checksums are disabled.
db_1   | 
db_1   | fixing permissions on existing directory /var/lib/postgresql/data ... ok
db_1   | creating subdirectories ... ok
db_1   | selecting default max_connections ... 100
db_1   | selecting default shared_buffers ... 128MB
db_1   | selecting default timezone ... UTC
db_1   | selecting dynamic shared memory implementation ... posix
[...]

PR #69, opened 2020-01-29
fix(tests): fix performance tests database url

madflow:
* assuming graphile_worker_perftest should be the suffix.
* probably related to #68

### DIFF
--- a/perfTest/run.sh
+++ b/perfTest/run.sh
@@ -7,7 +7,7 @@ cd "${0%/*}"
 export NO_LOG_SUCCESS=1
 
 # if connection string not provided, assume postgres is available locally
-export PERF_DATABASE_URL=${TEST_CONNECTION_STRING-graphile_worker_perftest}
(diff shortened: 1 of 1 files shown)


perfTest/run.sh
```diff
@@ -7,7 +7,7 @@ cd "${0%/*}"
 export NO_LOG_SUCCESS=1
 
 # if connection string not provided, assume postgres is available locally
-export PERF_DATABASE_URL=${TEST_CONNECTION_STRING-graphile_worker_perftest}
+export PERF_DATABASE_URL=${TEST_CONNECTION_STRING}-graphile_worker_perftest
```
benjie: This isn't the intention of this code; it's using bash parameter substitution to apply a default value if TEST_CONNECTION_STRING is not set. It seems, however, that TEST_CONNECTION_STRING is set, and that it's in use by something else so cannot be reset? I'll have to look into this more; I believe this was caused by #64

madflow: Ahh - Okay! I will close this PR then.

benjie: Thanks anyway!

PR #77, opened 2020-02-04
feat(perf): allow queue_name to be null

benjie:
Before:

```
Timing startup/shutdown time...
... it took 67ms

Scheduling 20000 jobs


Timing 20000 job execution...
Found 999!

... it took 3210ms
Jobs per second: 6363.16


Testing latency...
[core] INFO: Worker connected and looking for jobs... (task names: 'latency')
Beginning latency test
Latencies - min: 2.47ms, max: 12.00ms, avg: 2.72ms
Done
Done in 10.52s.
```

After:

```
Timing startup/shutdown time...
... it took 65ms

Scheduling 20000 jobs


Timing 20000 job execution...
Found 999!

... it took 2033ms
Jobs per second: 10165.04


Testing latency...
[core] INFO: Worker connected and looking for jobs... (task names: 'latency')
Beginning latency test
Latencies - min: 2.42ms, max: 12.37ms, avg: 2.71ms
Done
Done in 9.03s.
```

### DIFF
--- a/__tests__/main.runTaskListOnce.test.ts
+++ b/__tests__/main.runTaskListOnce.test.ts
@@ -12,7 +12,7 @@ test("runs jobs", () =>
     // Schedule a job
     const start = new Date();
     await pgClient.query(
-      `select * from graphile_worker.add_job('job1', '{"a": 1}')`
+      `select * from graphile_worker.add_job('job1', '{"a": 1}', queue_name := 'myqueue')`
     );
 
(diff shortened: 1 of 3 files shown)


benjie:
@gregplaysguitar Your suspicion about only using `job_queues` when necessary has panned out: we now max out at 10kjps versus 6.3kjps previously.

gregplaysguitar:
Nice!

PR #79, opened 2020-02-06
Add new and old json to the job invoke function

Ankcorn:
When I used postgraphile worker at a hackathon this one time it took me a little while to figure this out as my SQL knowledge is pretty bad. It ended up being really simple but this could help people when they want to trigger more complicated jobs :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -677,7 +677,9 @@ BEGIN
     'schema', TG_TABLE_SCHEMA,
     'table', TG_TABLE_NAME,
     'op', TG_OP,
-    'id', (CASE WHEN TG_OP = 'DELETE' THEN OLD.id ELSE NEW.id END)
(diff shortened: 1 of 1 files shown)


benjie:
I don't think this is what I'd recommend for most people - it inflates the jobs table and normally isn't required. Normally writing the ID is sufficient and then the job can look up the latest value later. If you wanted to add this as a separate code block with an explanation of when you'd use it that'd be welcome though :+1:

Ankcorn:
Ooh that makes sense. I have been spoilt by dynamodb streams. I did not realise it would increase the amount of data stored in my postgres database. Too used to stuff not having a cost to implement 😅

Thanks for lifting the hood on how postgraphile worker works for me🤘

benjie:
No problem. Basically it gets written to the `payload` column in the database; that JSON object would be larger. It's not a huge deal unless one of your columns is big.

PR #82, opened 2020-02-13
added completeJob and failJob to makeWorkerUtils

stlbucket:

PR #83, opened 2020-02-14
feat(admin): introduce admin commands

benjie:
Manipulating the underlying jobs/job_queues tables is Risky Business :tm:, so if you want to build a UI or administration interface over Graphile Worker it's best to use published APIs. But... We don't have many, at least not suitable ones. `fail_job` and `complete_job` are intended to be called by the worker itself _after_ it has retrieved a job via `get_job`; they are not suitable for general usage (which is why we closed #82).

This PR introduces some admin commands suitable for manipulating a Graphile Worker job queue. Specifically, it adds the `completeJobs`, `permanentlyFailJobs` and `rescheduleJobs` commands, all of which accept as an argument the list of job IDs to affect, and all of which perform the necessary consistency checks. One important thing to note is that these functions resolving positively does NOT guarantee all these jobs were updated - if another worker has locked a job then it will not be manipulated by these functions, but they will still resolve correctly - to check what has happened you should refetch the tasks after executing the methods.

I wanted to open this up to discussion before merging (also add tests), I know at least @stlbucket is interested in this topic.

Let me know if there's more functionality you'd like implemented that your administrative interface may depend upon.


benjie:
To pump up the excitement, here's a screenshot from @stlbucket in #82:

image

stlbucket:
to be clear, my original request was just about exposing completeJob and failJob via makeWorkerUtils.   i should have realized i don't need workerUtils for this UI as i have postgraphile exposing the graphile_worker schema so i can call them there.

that being said, whatever the proper methods are to accomplish this, i will incorporate into this ui.

the above is currently part of another project but i intend to pop it out into a standalone ui component soon.

garcianavalon:
All those methods are ver useful for our use case too, looking forward to them

benjie:
@garcianavalon Are there any more methods you need, or slightly different takes on these methods? Is the proposed "success even if it skips a few of the IDs" pattern okay for you? Also, what is your use-case?

garcianavalon:
"Pause" a job is something I've used in other job queues before, but if I recall correctly we already discussed this one and decided that rescheduling to a date far in the future was a better approach (thats covered by `rescheduleJobs`). Another useful one could be editing the payload of the job, but again this can be covered by a combination of `permanentlyFailJobs` and a new `addJob`

I like the proposed approach, specially if there are going to be conflicts often due to locking of jobs by workers. From a DX point of view, if the methods returned a summary of what happened to each job would be great. That would simplify reporting back the user on the UI, instead of having to compare the initial state vs the resulting one to see "what happened". For example:
```js
const result = await completeJobs([1, 2])
// and result contains something like
[ 'completed', 'locked, not modified']
```

Our use case for Graphile Worker is mainly running tasks triggered by user events (mail notifications on completion of tasks, distance matrix calculations between locations, clean up of bucket files, you get the idea) but we also have some recurring tasks (call X API every Y minutes and store the result). We have plans to build a UI for our team to monitor and manage this tasks (have they run correctly? do we need to manually retry something? etc.) as know we have to resort to "check the logs" or the database from time to time. A very common thing is that a job will be misconfigured (wrong our faulty payload) or depends on another API that might be down, so we need to pause/reschedule them or fail them and re-trigger with the correct parameters.

Right now we have delayed building this UI because a) we have a ton more stuff to do b) we wanted to give a little more time for this APIs to be developed. Looking forward to @stlbucket component too

benjie:
> From a DX point of view, if the methods returned a summary of what happened to each job would be great. That would simplify reporting back the user on the UI, instead of having to compare the initial state vs the resulting one to see "what happened". For example:

This is a great point. Rather than doing this, I've changed all the functions to `returning *` - you can assume that if you listed an id and it's not in the result set then it wasn't modified, but the reason for that is up to you to determine (we can't figure that out a way that doesn't compromise performance).

garcianavalon:
> > From a DX point of view, if the methods returned a summary of what happened to each job would be great. That would simplify reporting back the user on the UI, instead of having to compare the initial state vs the resulting one to see "what happened". For example:
> 
> This is a great point. Rather than doing this, I've changed all the functions to `returning *` - you can assume that if you listed an id and it's not in the result set then it wasn't modified, but the reason for that is up to you to determine (we can't figure that out a way that doesn't compromise performance).

That should be good enough, thanks for the addition

PR #89, opened 2020-02-28
Make worker schema configurable

### LINKED ISSUE #65: Make worker schema configurable
It would be great if the default `graphile_worker` schema could be made configurable (perhaps in the existing `src/config.ts`). It's a small and unimportant request, but in large projects with many other schemas in the database, it's nice to give it a name that follows the project naming convention.

benjie:
Fixes #65 (fao @demianuco)

~~Currently this is only configurable via `GRAPHILE_WORKER_SCHEMA` envvar, but we could/should add `cosmiconfig` or similar for the settings defaults.~~

### :sparkles: Features

- Can override the default `graphile_worker` schema of your choice using the `schema` option, `GRAPHILE_WORKER_SCHEMA` envvar or `schema` config
- Cosmiconfig added for overriding defaults for `pollInterval` or `concurrentJobs`

### :rotating_light: Changed API signatures:

NOTE: if you use the CLI then you are unaffected by this change; only library users will be affected.

Overriding the SQL schema to use means that everything in the codebase needs to know this setting. To achieve this:

- all major APIs now accept `options` as a configuration parameter
- where this was optional before it is now required
- where options was not the first argument, it has been moved to the first argument (for consistency)

As such the following APIs (most of which are internal) have been changed:

- `getTasks(taskPath, watch, logger)` -> `getTasks(options, taskPath, watch)`
- `runTaskList(tasks, pgPool, options?)` -> `runTaskList(options, tasks, pgPool)`
- `runTaskListOnce(tasks, client, options?)` -> `runTaskList(options, tasks, client)`
- `migrate(client)` -> `migrate(options, client)`
- `makeAddJob(withPgClient)` -> `makeAddJob(options, withPgClient)`
- `makeJobHelpers(job, { withPgClient }, baseLogger)` -> `makeJobHelpers(options, job, { withPgClient, logger? })`
- `makeNewWorker(tasks, withPgClient, options, continuous)` -> `makeNewWorker(options, tasks, withPgClient, continuous)`

Also if you're a TypeScript user: we've renamed `WorkerSharedOptions` to `SharedOptions` and added a new `WorkerSharedOptions`. This is particularly relevant if you're using the WorkerUtils class.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,7 @@ jobs:
           node-version: ${{ matrix.node-version }}
       - run: yarn --frozen-lockfile
       - run: yarn lint
-      - run: yarn jest -i
+      - run: yarn jest -i --ci
 
   altschema:
     runs-on: ubuntu-18.04
@@ -69,4 +69,4 @@ jobs:
           node-version: 12.x
       - run: yarn --frozen-lockfile
       - run: yarn lint
-      - run: yarn jest -i
+      - run: yarn jest -i --ci
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,3 +41,32 @@ jobs:
       - run: yarn --frozen-lockfile
       - run: yarn lint
       - run: yarn jest -i
+
+  altschema:
+    runs-on: ubuntu-18.04
+    env:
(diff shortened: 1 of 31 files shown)


benjie:
Before merging this I'd like to refactor the entire API so that everything takes `options` as the first argument.

benjie:
A lot more work has been done on this; please see updated description above.

PR #95, opened 2020-03-10
fix: use default message when task rejects with non-Error object

parker-torii:
In the case that a worker task throws a poorly formed error (one without a `message` property), the error message prints but the query that would fail the job does not complete.  This leaves the job not running but still occupying the queue.

### DIFF
--- a/src/worker.ts
+++ b/src/worker.ts
@@ -168,11 +168,20 @@ export function makeNewWorker(
       const durationRaw = process.hrtime(startTimestamp);
       const duration = durationRaw[0] * 1e3 + durationRaw[1] * 1e-6;
       if (err) {
-        const { message, stack } = err;
+        const { message: rawMessage, stack } = err;
+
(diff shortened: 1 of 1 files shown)


src/worker.ts
```diff
@@ -185,7 +185,7 @@ export function makeNewWorker(
         await withPgClient(client =>
           client.query({
             text: `SELECT FROM ${escapedWorkerSchema}.fail_job($1, $2, $3);`,
-            values: [workerId, job.id, message],
+            values: [workerId, job.id, message || ''],
```
benjie: Thanks for spotting this. Let's fix it in the `logger.error` call too. I can't suggest the change up there, but here's the logic I'm after

```ts
const { message: rawMessage, stack } = err;
const message = rawMessage || String(err) || 'Non-error thrown (no message)';
```

Then replace `err.message` with `message`.

Issue #85, opened 2020-02-19
Make an official Docker image


benjie:
Preferably with instructions about how to add your tasks.

keepitsimple:
Hi @benjie. Can I make a PR with small polishing for the docker & docker compose?

benjie:
That’d be ace 👍

keepitsimple:
@benjie I cannot push a new branch and make a PR. Can you advice please?

benjie:
Please fork the repo and make a PR from there; see: https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request-from-a-fork

You can use your existing clone to do so; after you've created your fork you should use `git remote` to update your remote's URL; e.g.:

```
git remote set-url origin git@github.com:keepitsimple/worker.git
```

then you can `git push` and raise a PR from the GitHub interface on your fork.

benjie:
Hey @keepitsimple how's it going?

benjie:
This was implemented in https://github.com/graphile/worker/pull/111

PR #111, opened 2020-04-28
docker: official Docker image

madflow:
#85 

* Adapted multi-stage commits from Postgraphile
* Added some docs
* Tasks are added via volume mount


benjie:
Docker hub is configured, but the builds are still progressing. Will check back later.

benjie:
Can confirmed this works and all the auto-builds are working too. Thanks @madflow :raised_hands:

PR #115, opened 2020-05-27
Tweak schema exist check logic

joshbwlng:
A small tweak to the way the schema exist checks operates. Instead of assuming the schema exists and catching when it doesn't, this change will execute a `SELECT EXISTS` statement to check existence. The current `try/catch` logic ensures that Postgres will have error output at least once causing confusion for those using this library.

Closes https://github.com/graphile/worker/issues/116

### DIFF
--- a/src/migrate.ts
+++ b/src/migrate.ts
@@ -53,21 +53,24 @@ export async function migrate(
 ) {
   const { escapedWorkerSchema } = processSharedOptions(options);
   let latestMigration: number | null = null;
-  try {
-    const {
-      rows: [row],
-    } = await client.query(
-      `select id from ${escapedWorkerSchema}.migrations order by id desc limit 1;`,
-    );
-    if (row) {
-      latestMigration = row.id;
-    }
-  } catch (e) {
-    if (e.code === "42P01") {
-      await installSchema(options, client);
-    } else {
-      throw e;
-    }
+
+  const {
+    rows: [schemaExists],
+  } = await client.query({
+    text: `SELECT EXISTS(SELECT 1 FROM information_schema.schemata WHERE schema_name=$1);`,
+    values: [escapedWorkerSchema],
+  });
+  if (!schemaExists.exists) {
+    await installSchema(options, client);
+  }
+
+  const {
+    rows: [row],
+  } = await client.query(
+    `select id from ${escapedWorkerSchema}.migrations order by id desc limit 1;`,
+  );
+  if (row) {
+    latestMigration = row.id;
   }
   const migrationFiles = (await readdir(`${__dirname}/../sql`))
     .filter((f) => f.match(/^[0-9]{6}\.sql$/))
--- a/src/migrate.ts
+++ b/src/migrate.ts
@@ -55,12 +55,12 @@ export async function migrate(
   let latestMigration: number | null = null;
 
   const {
-    rows: [schemaExists],
-  } = await client.query({
-    text: `SELECT EXISTS(SELECT 1 FROM information_schema.schemata WHERE schema_name=$1);`,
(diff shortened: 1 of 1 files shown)


src/migrate.ts
```diff
@@ -53,21 +53,24 @@ export async function migrate(
...
+
+  const {
+    rows: [schemaExists],
+  } = await client.query({
+    text: `SELECT EXISTS(SELECT 1 FROM information_schema.schemata WHERE schema_name=$1);`,
+    values: [escapedWorkerSchema],
+  });
+  if (!schemaExists.exists) {
+    await installSchema(options, client);
+  }
```
benjie, line 65: It's more performant to query pg_namespace than information_schema; I've also edited the code to be more consistent with the rest of the codebase. You shouldn't use `escapedWorkerSchema` in a value position because it's escaped as an identifier (so for `MySchema` it'd have the value `"MySchema"` (with quotes)), instead use the `workerSchema` value directly.

```suggestion
  const {
    rows: [pgNamespace],
  } = await client.query(
    `select oid from pg_catalog.pg_namespace where nspname = $1;`,
    [workerSchema],
  );
  if (!pgNamespace) {
    await installSchema(options, client);
  }
```

benjie:
The tests are failing (this change seems to be causing the schema to be attempted to be created on every migration) - I suspect this is because you're using `escapedWorkerSchema` in a placeholder.

joshbwlng:
> I'm hesitant to add another database roundtrip for this (hence the try/catch approach) since it'll only be needed once per database. Sending this additional query every single time you run worker for the next few years just so that no error is raised the first time seems inefficient.
> 
> Can you confirm _where_ you're seeing this error? Graphile Worker itself does not output it (you can see the `catch` deliberately swallows the error) - is it perhaps in your Postgres logs?

@benjie 
Thanks for taking a look at this.
Sorry if my description wasn't quite descriptive enough, but yes, these are found in Postgres logs. I understand the argument of it being less efficient, but it seems odd to write code that ensures that a database error will occur at least once. If it was a conscious decision to allow this error in order to gain performance, then this PR can be closed. I can walk away knowing its a part of the design and not an oversight.

benjie:
> I understand the argument of it being less efficient, but it seems odd to write code that ensures that a database error will occur at least once.

Yeah, I went back and forth on it before erring on the side of performance (as I so often do). Honestly I wasn't expecting anyone to notice the error in the Postgres logs.

> If it was a conscious decision to allow this error in order to gain performance, then this PR can be closed. I can walk away knowing its a part of the design and not an oversight.

It was, but that doesn't mean it's the best decision. If we can solve this without adding a database roundtrip I'd be happy.

I considered adding `create schema if not exists...` to the query but some people prefer to manage migrations separately such that worker doesn't need "create" permissions during runtime; and besides it makes the query slower so again I'd rather avoid it.

We could add a (small) comment to the statement, then the comment might be output as part of the error message and make more sense when you read through?

---

Having spent a couple minutes thinking about it, I'm happy with the compromise the current codebase makes with regard to this error, so I'm going to go ahead and close this issue. Thanks for the discussion/contribution though, and let me know if you come up with a better solution :+1:  Maybe just documenting the error in the README would be enough, so Google'd find it? :thinking:

joshbwlng:
@benjie I do think a quick mention about it in the README would be nice so people can find the cause of the error quickly when/if they notice it in their logs. We all use a lot of libraries and sometimes its hard to track these kinds of things down. :+1:

benjie:
Would you like to raise a PR? Add it just above the [development header](https://github.com/graphile/worker#development) and include an exact copy of the error along with information of where you might find it :+1:

Issue #118, opened 2020-06-27
Add multiple leaky bucket rate limiting


ben-pr-p:
I have a `send-message` job that sends a text message. We have a global 3,000 messages / minute rate limit, a limit of 6 messages / second per sending phone number, and have many different clients that send messages. Each client has hundreds of phone numbers.

Our goals are:
a) To stay compliant with our global rate limit
b) To stay compliant with our per-phone number rate limit
c) To prevent any client from clogging the queue for all other clients, such that one client sending 6,000 messages in a minute means that all other clients messages are delayed by 2. Something like 1000 / minute would probably be sensible here, given that not all clients are going to send their maximum volume at once.

One way to do this is a Postgres friendly simplification of the [leaky bucket algorithm](https://en.wikipedia.org/wiki/Leaky_bucket), where you would have buckets:
```
buckets (
    bucket_type text,
    interval text, // 'second' or 'minute'
    capacity integer
)
```
And bucket_intervals:
```
bucket_intervals (
    bucket_type text,
    bucket_name text,
    bucket_time timestamp, // truncated to the second or minute based on the bucket interval
    tokens_used integer, // number of jobs in this bucket
    primary key (bucket_time, bucket_type, bucket_name)
)
```

Whenever `run_at` is computed, if the job has specified buckets (via a `buckets text[]` column / parameter), run_at would be set to the max of the user specified run_at (or now) and the next available slot which doesn't overflow any bucket interval.

For our use case, our buckets table would be:
```
insert into buckets (bucket_type, interval, capacity)
values 
    ('send-message-global', 'minute', 3000),
    ('send-message-client', 'minute', 1000),
    ('send-message-phone', 'second', 6);
```

And each `send-message` job would be queued with three buckets:
1) `send-message-global` (which would map to `bucket_type = 'send-message-global', bucket_name = null`)
2) `send-message-client|<client-id>` (`bucket_type = 'send-message-client', bucket_name = <client-id>`)
3) `send-message-phone|<phone-number>` (`bucket_type = 'send-message'phone`, bucket_name = <phone-number`)

I think this could be accomplished via:
1) adding an optional buckets parameter to add_job
2) triggers on the `jobs` table that only run when the job has buckets, and as a result have no negative performance impact for users who don't need this feature

To keep it performant, we would need to delete past buckets. This could either be done on job completion / failure, or we could just write a SQL function to do it and leave periodically calling that function up to the user.

[...]

PR #128, opened 2020-08-16
Feat/forbidden flags

ben-pr-p:
This implements forbidden flags as per https://github.com/graphile/worker/issues/118.

As far as I can tell by running and rerunning the perfTest at various stages of development and at the end, this has no impact on performance for users who choose not to use the feature.

Right now, 14.9.6 is failing with a Postgresql syntax error - do you know if there is something funky with `pg` going on in that version?

### DIFF
--- a/src/worker.ts
+++ b/src/worker.ts
@@ -99,7 +99,7 @@ export function makeNewWorker(
           text:
             // TODO: breaking change; change this to more optimal:
             // `SELECT id, queue_name, task_identifier, payload FROM ${escapedWorkerSchema}.get_job($1, $2);`,
-            `SELECT * FROM ${escapedWorkerSchema}.get_job($1, $2, $3);`,
+            `SELECT * FROM ${escapedWorkerSchema}.get_job($1, $2, forbidden_flags => $3);`,
           values: [workerId, supportedTaskNames, flagsToSkip],
           name: noPreparedStatements ? undefined : `get_job/${workerSchema}`,
         }),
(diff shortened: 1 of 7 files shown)


src/worker.ts
```diff
@@ -77,15 +78,29 @@ export function makeNewWorker(
...
+      if (Array.isArray(forbiddenFlags)) {
+        flagsToSkip = forbiddenFlags;
+      } else if (typeof forbiddenFlags === "function") {
+        const forbiddenFlagsResult = forbiddenFlags();
+
+        if (Array.isArray(forbiddenFlagsResult)) {
+          flagsToSkip = forbiddenFlagsResult;
+        }
+
+        flagsToSkip = await forbiddenFlagsResult;
```
benjie, line 92:
```suggestion
        } else {
          flagsToSkip = await forbiddenFlagsResult;
        }
```

benjie:
> Right now, 14.9.6 is failing with a Postgresql syntax error - do you know if there is something funky with pg going on in that version?

Could it be that it's actually PostgreSQL 9.6 kicking out the syntax somewhere? The PR's a little too large for me to quickly scan it and spot the issue.

`pg` and Node 14 did have weird interactions for a while, but I've only seen it "hanging" rather than giving syntax errors.

I really need to track a version of the PostgreSQL schema in version control so I can track the actual diff.

(I don't have time to fully review this today, but if we can get this and #129 approved I'll merge them into a single migration manually later.)

ben-pr-p:
Closing in favor of https://github.com/graphile/worker/pull/131

Issue #129, opened 2020-08-17
Counter for how many times a task has been debounced


lukeramsden:
Proposal: a counter on each job that is incremented every time it has been debounced through the job key.

My use-case is I'd like to debounce notifying the user about messages, so of course I'm using the job 
key for that, but I'd like to have the notification be "You received X new message(s)" - the ideal solution to this would just be to have a field on the job that is incremented on every debounce, and then I just pass that to my email template.

benjie:
I can imagine a world in which I approved a "revision" or "version" counter column being added to a job. It would be updated on any change to the task identifier, queue or payload (but not run_at, attempts, last_error). Would that solve your needs?

lukeramsden:
I believe so. I'm happy to attempt a PR for this.

benjie:
This was implemented in #130

PR #130, opened 2020-08-17
(#129) Added 'revision' column that is incremented every time a job is debounced via job_key

lukeramsden:


sql/000005.sql
```diff
@@ -0,0 +1,105 @@
+alter table :GRAPHILE_WORKER_SCHEMA.jobs add column revision int default 0 not null;
```
lukeramsden, line 1: Should this start at 0 or 1?

benjie: 0 :+1:

benjie:
From a quick scan this looks great. I won't be able to fully review it for a little while.

PR #131, opened 2020-08-21
feat: forbidden flags

benjie:
This is a modified version of #128.

### DIFF
--- a/sql/000005.sql
+++ b/sql/000005.sql
@@ -1,5 +1,5 @@
 alter table :GRAPHILE_WORKER_SCHEMA.jobs add column revision int default 0 not null;
-alter table :GRAPHILE_WORKER_SCHEMA.jobs add column flags text[] default null; 
+alter table :GRAPHILE_WORKER_SCHEMA.jobs add column flags jsonb default null;
 
 drop function :GRAPHILE_WORKER_SCHEMA.add_job(text, json, text, timestamptz, int, text, int);
 create function :GRAPHILE_WORKER_SCHEMA.add_job(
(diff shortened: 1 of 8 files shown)


sql/000005.sql
```diff
@@ -94,11 +104,81 @@ begin
...
+        where job_queues.queue_name = jobs.queue_name
+        and (job_queues.locked_at is null or job_queues.locked_at < (v_now - job_expiry))
+        for update
+        skip locked
+      )
+    )
+    and run_at <= v_now
+    and attempts < max_attempts
+    and (task_identifiers is null or task_identifier = any(task_identifiers))
+    and (forbidden_flags is null or (flags ?| forbidden_flags) is not true)
```
blasterpistol, line 155: In my limited experience with performance optimizations, nullable arguments in clauses such as `where $1 is null or $1 ...` force the planner not to use the index, but I'm not 100% sure about that, I just remember that this variation worked faster `where $1 is not null and $1 ...`

blasterpistol: quick test on my own dataset
image
image

benjie: Thanks for taking the time to analyse this :) The conditions you've displayed are not equivalent. The first would return a result where `phone` is null (desired), the latter will not. We're broadly dealing with the situation where this column would be null in most cases, this is not expected to be an index concern (the field isn't indexed).

The syntax avoiding `OR` for `(phone is null) or (phone like '1111111')` would be `not ((phone is not null) and (phone not like '1111111'))`, and maybe that would be faster if we were indexing this field as it would allow excluding rows where there was an index match; I'm not sure though - it's a pretty broad condition to say `phone not like '1111111'`, I suspect PostgreSQL would opt for a table scan?

benjie:
I just realised I let you get away without editing the README @ben-pr-p :laughing:

ben-pr-p:
Another PR?

benjie:
If you'd be so kind :heart: :+1:

Issue #57, opened 2019-12-18
Allow TS users to overload `addJob` (take two)


keithlayne:
It would be nice for TS users who enqueue jobs from within tasks (or via the `addJob` function returned by `run`) to have payload types checked at the call site.

**NOTE:** This has nothing to do with validating payloads.  This feature would be only useful insofar as the payload types are correct, and would only provide type checking at `addJob` call sites.

In my current setup, I use `runtypes` to derive both payload validators and the static types that I would use in these overloads.  In this way there is a single source of truth, and when the runtime type is updated, TS will trickle that down to all call sites.  This feels pretty robust.

Adding this feature would give additional type checks and an improved editor experience.

The requirement for this to work is that you pass a string literal (or a value typed as a non-union string literal) as the first parameter.  It's assumed that if you add an entry for a task that the payload will be a required argument.  You can however make a payload type nullable if you wish.

Passing any value that isn't inferred as a string key of the `TaskPayload` should result in falling back to the original function type.  The reason for the non-union key requirement is that in that case (with current implementation) you could satisfy the type checker with an incorrect payload type.

This is (almost) fully implemented in the playground link below:

[...]

PR #142, opened 2020-09-09
Add generics for typed payloads

warent:
Payloads for jobs are untyped and require unnecessary/unsafe casting, e.g

```ts
type MyType = {
  Val: string;
}
const task: Task = async (payload) => {
  const { Val } = payload as MyType;
  // etc
}
```

This PR enables us to utilize TS generics:

```ts
type MyType = {
  Val: string;
}
const task: Task<MyType> = async ({ Val }) => {
  // etc
}
```

Note: this update is backward compatible i.e. does not require changes to existing code since the generic is optional.

### DIFF
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -139,8 +139,8 @@ export interface WorkerUtils extends Helpers {
   ) => Promise<Job[]>;
 }
 
-export type Task = (
-  payload: unknown,
+export type Task<T = unknown> = (
+  payload: T,
   helpers: JobHelpers,
 ) => void | Promise<void>;
 
@@ -160,11 +160,11 @@ export interface WatchedTaskList {
   release: () => void;
 }
 
-export interface Job {
+export interface Job<T = unknown> {
   id: string;
   queue_name: string | null;
   task_identifier: string;
-  payload: unknown;
+  payload: T;
   priority: number;
   run_at: Date;
   attempts: number;


benjie:
Hi @warent; thanks for your interest in contributing to Graphile Worker! Using TypeScript types like this can be misleading since jobs can be (and often are) added outside of a TypeScript context; so there's no guarantee that the payloads actually conform to the TS types defined here. This can lead to you assuming that something is a number when actually it's `null`, resulting in more bugs in your code.

I recommend using a type guard or assertion; e.g.

```ts
interface MyPayload {
  username: string;
}

function assertMyPayload(payload: any): asserts payload is MyPayload {
  if (typeof payload === 'object' && payload && typeof payload.username === 'string') {
    return;
  }
  throw new Error("Invalid payload, expected a MyPayload");
}

const task: Task = async (payload) => {
  assertMyPayload(payload);
  console.log(payload.username);
}
```

If this is too manual, you might prefer to use a library such as `runtypes` or the many others of a similar ilk. If you're not concerned with the type safety of the payload, you can work around it with a couple casts:

```ts
const task: Task = inPayload => {
  const payload = (inPayload as any) as MyPayload;
}
```

We discussed a cohesive typing solution in depth in #57, whatever solution we added we'd want it to work for `addJob` too, giving at least slightly stronger guarantees the types would be respected.

PR #144, opened 2020-09-23
docs: add note regarding the required db dump step

gregplaysguitar:
This is a little tricky if you don’t have the right pg version running already


benjie (approved):
Awesome; thanks!

benjie:
At some point I should split that section into a CONTRIBUTING file.

PR #146, opened 2020-09-29
fix(watch): solve bug with cyclic requires

### LINKED ISSUE #145: Watch mode triggers 'Maximum call stack size exceeded' on tasks with cyclic dependencies
### Summary

when running in watch mode a task with a dependency that is cyclic causes a stack overflow due to the reload method.

### Steps to reproduce

to follow

benjie:
Fixes #145; the module cache was being written to at too late a stage, causing cyclic requires to trigger an infinite loop. This PR fixes this by caching the module _before_ we execute the module, meaning the cache is in place before the module can require other modules which might in turn require this module again.

(We fixed this on stream.)

### DIFF
--- a/src/module.ts
+++ b/src/module.ts
@@ -74,12 +74,12 @@ ${code};
   // Since the code is evaluated synchronously, we set this immediately before
   // compiling and clear after.
   global["graphileWorker_fauxRequireCache"] = cache;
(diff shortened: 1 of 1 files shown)


hiddenboox:
Thanks @benjie for this fix! btw this lib is great

PR #147, opened 2020-10-26
feat(perf): extend index with `locked_at` and make partial

ben-pr-p:
## Description

Adds an index to improve the performance of job discovery when there are a few thousand failed or locked jobs.

I'm not sure if this is a change that is desirable for all users, but it is for us so I thought it would be at least worth opening a PR (even if it will just be kept closed) so that others who may have the problem can find this solution.

## Performance impact

On my machine (Macbook Pro 2017, 2.8 GHz Quad-Core Intel Core i7), this has no impact on `yarn perfTest`.

In production on several deployments, this dramatically speeds up `graphile_worker.get_job` when the number of jobs in a permanently failed state or locked / abandoned state increases above a few thousand. When this happens, `graphile_worker.get_job` becomes responsible for 90% of database CPU on a large (30 vCPU) database sequentially scanning over the same few thousand failed rows again and again.

## Security impact

None

## Checklist

### DIFF
--- a/sql/000006.sql
+++ b/sql/000006.sql
@@ -1,0 +1,6 @@
+create index jobs_priority_run_at_id_locked_at_without_failures_idx
+  on :GRAPHILE_WORKER_SCHEMA.jobs (priority, run_at, id, locked_at)
+  where attempts < max_attempts;
+
+drop index :GRAPHILE_WORKER_SCHEMA.jobs_priority_run_at_id_idx;
+
--- a/__tests__/migrate.test.ts
+++ b/__tests__/migrate.test.ts
@@ -33,7 +33,7 @@ test("migration installs schema; second migration does no harm", async () => {
     const { rows: migrationRows } = await pgClient.query(
       `select * from ${ESCAPED_GRAPHILE_WORKER_SCHEMA}.migrations`,
     );
-    expect(migrationRows).toHaveLength(5);
+    expect(migrationRows).toHaveLength(6);
     const migration = migrationRows[0];
     expect(migration.id).toEqual(1);
 
--- a/__tests__/schema.sql
+++ b/__tests__/schema.sql
@@ -355,7 +355,7 @@ ALTER TABLE ONLY graphile_worker.jobs
     ADD CONSTRAINT jobs_pkey PRIMARY KEY (id);
 ALTER TABLE ONLY graphile_worker.migrations
     ADD CONSTRAINT migrations_pkey PRIMARY KEY (id);
-CREATE INDEX jobs_priority_run_at_id_idx ON graphile_worker.jobs USING btree (priority, run_at, id);
+CREATE INDEX jobs_priority_run_at_id_locked_at_without_failures_idx ON graphile_worker.jobs USING btree (priority, run_at, id, locked_at) WHERE (attempts < max_attempts);
 CREATE TRIGGER _100_timestamps BEFORE UPDATE ON graphile_worker.jobs FOR EACH ROW EXECUTE PROCEDURE graphile_worker.tg__update_timestamp();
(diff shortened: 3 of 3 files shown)


benjie:
I wonder if `locked_at` should be the first field in the index? I've not checked over the code to validate this gut feeling.

benjie:
(I also wonder if using a partial index instead might be a good solution; but I'm not sure of the cost of removing/adding items to the index so frequently.)

ben-pr-p:
> (I also wonder if using a partial index instead might be a good solution; but I'm not sure of the cost of removing/adding items to the index so frequently.)

Do you mean a partial index on locked_at? The issue with that is that we actually query locked_at for the expiry check.



> I wonder if `locked_at` should be the first field in the index? I've not checked over the code to validate this gut feeling.

I think if the index order doesn't match the sort order, it can't use the index for the sorting, and then you end up with a sort over all non locked rows (expensive).

benjie:
> Do you mean a partial index on locked_at? The issue with that is that we actually query locked_at for the expiry check.

I was thinking more about the exhausted tries/failed forever. But actually that number is dynamic anyway, so I don't think that's tenable. I wonder if we should migrate jobs to a different table when they fail forever.

> I think if the index order doesn't match the sort order, it can't use the index for the sorting, and then you end up with a sort over all non locked rows (expensive).

Yeah, I was hoping the filtering would come into play first, but generally jobs shouldn't be locked, so that should be something we check very late. Though I'm not sure that it shouldn't come before `id`.

johnhaley81:
>  I wonder if we should migrate jobs to a different table when they fail forever.

I feel like that makes the most sense instead. The `jobs` table should be as high performance as possible (IMO) and adding indexes to allow for the table to grow really large will have a big effect on the performance of the system as a whole.

sql/000006.sql
```diff
@@ -0,0 +1,6 @@
+create index jobs_priority_run_at_id_locked_at_without_failures_idx
+  on :GRAPHILE_WORKER_SCHEMA.jobs (priority, run_at, id, locked_at)
+  where attempts < max_attempts;
```
benjie, line 3: Given `locked_at` is after `id` in the index (which is unique), I feel like it's unlikely to give much value adding it to the index, but it will increase both churn on the index and the size of the index. How about we just do the partial index thing; does that still improve performance in your use case? It'll actually reduce index size which is a great addition; this could really make a difference in the case of lots of perma-fails.

```suggestion
create index jobs_priority_run_at_id_alive_idx
  on :GRAPHILE_WORKER_SCHEMA.jobs (priority, run_at, id)
  where attempts < max_attempts;
```

benjie (changes requested):
Let's drop locked_at from the index if that still solves your performance issue. Then this is a very obvious merge :+1:

benjie:
@ben-pr-p If you drop `locked_at` from the index does it still fix your performance issues?

ben-pr-p:
> @ben-pr-p If you drop `locked_at` from the index does it still fix your performance issues?

I can milk a bit exta out of including `locked_at`, specifically at the end in most cases, but I'm also not benchmarking update time. I'm fine with dropping it

benjie:
If we didn't support PG10, we could do `CREATE INDEX ... INCLUDE (locked_at)`; but alas that was only added in PG11.

benjie (approved):

benjie:
Since this slightly expands the size of rows in the index, but overall many decrease the size of the index (because it doesn't index expired jobs, which build up over time), I'm going to merge this. We should solve it better in future, but this is a very simple solution.

benjie:
Thanks @ben-pr-p  :raised_hands:

ben-pr-p:
> If we didn't support PG10, we could do `CREATE INDEX ... INCLUDE (locked_at)`; but alas that was only added in PG11.

I'd be interested in the difference in update time with `include` vs as the last parameter for different workloads, just as an aside. I've seen the include option in the docs but I have no idea whether / when to actually use it!

Issue #56, opened 2019-12-13
[Discussion] Let TS users overload `addJob`


keithlayne:
I see two different options for doing this, and there are tradeoffs for each.  Wanted to lay those out here before implementing (none entail much work, and I don't *think* any would be breaking changes):

*NOTE:* For any of these to work (assuming they are in `interfaces.ts`, the interface being augmented should to be exported by name from the top level.

### Option 1: Change `AddJobFunction` to an interface with a call signature

Changes:
```ts
export interface AddJobFunction {
  (identifier: string, payload?: any, options?: TaskOptions): Promise<Job>;
}
```
Adding an overload:
```ts
import { Job, TaskOptions } from 'graphile-worker'

declare module 'graphile-worker' {
 interface AddJobFunction {
    (identifier: 'taskName', payload: Payload, options?: TaskOptions): Promise<Job>
  }
}
```
Special TS rules bump this overload to the top of the list due the non-union literal.

Pros:
* Users can't screw it up at call sites

Cons:
* Verbose
* Have to import extra types

### Option 2: Generic overload with a type map

Changes:
```ts
export interface TaskPayload {}
export interface AddJobFunction {
  <T extends keyof TaskPayload>(identifier: T, payload: TaskPayload[T], options?: TaskOptions): Promise<Job>
  (identifier: string, payload?: any, options?: TaskOptions): Promise<Job>;
}
```
Adding an overload:
```ts
declare module 'graphile-worker' {
 interface TaskPayload {
    taskName: Payload
  }
}
```

Pros:
* Easier to add an "overload"

Cons:
* Users can pass the type param explicitly or pass something other than a string literal as the task name; this could cause bogus inputs to type check

I had another idea that actually won't work because the payload is required in (at least most) overload signatures and not in the default signature.

I started writing this favoring option 2, but at this point I feel like correctness trumps brevity.  They're both easy to implement, so whatever is decided here can be turned around quickly.

keithlayne:
The limitation on the second option could probably be overcome with some kind of `NonUnion` type.  Not sure if that's worth it, but I do like the improved ergonomics of that option.

keithlayne:
[...]

PR #150, opened 2020-11-14
feat(types): allow configuring task payloads

joelmukuthu:
## Description

This PR makes the `Task` type generic, thereby allowing users to configure the exact shape of payloads.

## Performance impact

N/A

## Security impact

N/A

## Checklist

- [x] ~~I've added tests for the new feature, and `yarn test` passes~~
- [x] ~~If this is a breaking change I've explained why.~~

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -3,6 +3,7 @@
 ### Pending
 
 - Add your changes here
+- TypeScript: make the `Task` type generic, to allow specifying payloads
 
 ### v0.8.1
 
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -139,8 +139,8 @@ export interface WorkerUtils extends Helpers {
   ) => Promise<Job[]>;
 }
 
-export type Task = (
-  payload: unknown,
+export type Task<Payload extends unknown = unknown> = (
+  payload: Payload,
   helpers: JobHelpers,
 ) => void | Promise<void>;
 


benjie:
Thanks for the pull request, it seems to be an alternate implementation of #142. I'm not really happy adding this to tasks if there's no synchronised type definitions with addJob - otherwise there's not really any type safety here. Further there's no guarantee that the tasks are added with addJob, they might be created in the database directly, so any types applied to the payload cannot be trusted, hence why we currently type them as unknown. Basically - the pain of `unknown` is intentional so that you know you shouldn't be trusting the data you get from payload because it's not guaranteed to conform to your type definition. Currently we recommend you use `runtypes` or similar approaches to ensure that the payload does conform to the shape you're expecting before you go ahead and use it.

For previous work and discussions around this, please see:

- https://github.com/graphile/worker/issues/56
- https://github.com/graphile/worker/issues/57
- https://github.com/graphile/worker/pull/142

joelmukuthu:
In my case the tasks are created in the database (which is why I missed the `addJob` use-case) and I would just like to document the shape of the payload in my TS code and of course have some type-checking there as well. It would be nice if this was opt-in -- with responsibility falling to the user once they opt in -- and I would also argue that while the shape of payloads is unknown to Graphile worker, it should be known to the user.

That said, I understand your reservations and motivation to protect users from foot-guns and that it's a non-trivial issue. So it's fine by me to close this :)

benjie:
I'm kind of on the fence. I'm going to close this because we already have issues/PRs covering this topic, but I appreciate you adding your +1. I wouldn't be surprised if I give in and add types in the next 12 months or so :shrug:

joelmukuthu:
No worries! Thank you so much for your continued efforts on OSS! :heart:

PR #154, opened 2020-12-02
feat(job_key): allow controlling job_key's updates

benjie:
## Description

Currently `job_key` performs "debouncing" - an existing job is updated to have all the new values, including `run_at`. This is useful in many circumstances, however sometimes users want more of a "throttling" approach where the job is guaranteed to be executed at the original `run_at`, but further calls before it executes will not cause multiple executions. Other times users might want full "de-duplication" where the new job will not be added if a previous matching job has ever been scheduled (we cannot fully achieve this since completed jobs are deleted, so there's no permanent record of their job key - if you need this you should use a table you control to track these values and ensure uniqueness).

This PR introduces the `job_key_mode` parameter which allows you to choose what to do in the event that a job with matching job key is found. When reviewing this PR it's suggested that you read the `README` first to familiarise yourself with the intent of this PR. The default behaviour (`replace`) is the same as it was before, but we introduce two new modes: `preserve_run_at` which can be used for throttling, and `unsafe_dedupe` which will not trigger a new run when there's an existing job even if the existing job is in progress or has failed forever.

Further this PR "fixes" remove_job - if a job is in progress then it will have its `attempts` set to `max_attempts` such that on failure it will not be re-attempted, and it will have its key cleared such that a future `add_job` using the same key will be unaffected by the existing record.

## Performance impact

Benchmarking doesn't show any measurable difference in performance.

## Security impact

None known.


README.md
```diff
@@ -461,9 +461,12 @@ The `addJob` arguments are as follows:
...
   - `maxAttempts`: how many retries should this task get? (Default: 25)
-  - `jobKey`: unique identifier for the job, used to update or remove it later
-    if needed (see [Updating and removing jobs](#updating-and-removing-jobs));
-    can also be used for de-duplication
+  - `jobKey`: unique identifier for the job, used to replace, update or remove
+    it later if needed (see
+    [Updating and removing jobs](#updating-and-removing-jobs)); can be used for
+    de-duplication (i.e. throttling or debouncing)
+  - `jobKeyMode`: controls the behavior of `jobKey` when a matching job is found
+    (see [Updating and removing jobs](#updating-and-removing-jobs))
```
freakst, line 469: Internal heading #link won't work, because of renamed heading (same issue above).
```suggestion
    (see [Replacing, updating and removing jobs](#replacing-updating-and-removing-jobs))
```

README.md
```diff
@@ -461,9 +461,12 @@ The `addJob` arguments are as follows:
   - `queueName`: the queue to run this task under
   - `runAt`: a Date to schedule this task to run in the future
   - `maxAttempts`: how many retries should this task get? (Default: 25)
-  - `jobKey`: unique identifier for the job, used to update or remove it later
-    if needed (see [Updating and removing jobs](#updating-and-removing-jobs));
-    can also be used for de-duplication
+  - `jobKey`: unique identifier for the job, used to replace, update or remove
+    it later if needed (see
+    [Updating and removing jobs](#updating-and-removing-jobs)); can be used for
```
freakst, line 466:
```suggestion
    [Replacing, updating and removing jobs](##replacing-updating-and-removing-jobs)); can be used for
```

benjie:
Great catch @freakst :+1:

README.md
```diff
@@ -803,15 +823,47 @@ CREATE TRIGGER generate_pdf_update
...
+scheduled at a time, or to update other settings for the job. Behavior of this
+updating is controlled by the `job_key_mode` setting, but in all cases the
+following is true:
+
+- If an existing job with the same job key is found:
+  - If the existing job is locked:
+    - it will have it's `key` cleared
+    - it will have its attempts set to `max_attempts` to avoid it running again
+    - a new job will be created with the given attributes.
+  - Otherwise, if the existing job has previously failed:
```
freakst, line 842: As a clue: this approach imo potentially complicates tracking failed jobs in a pipe as job as an entity, identified by some unique id can suddenly transfer from state with non-zero amount of fails to straight successful and confuse metrics (specifically calculating failed jobs and job errors). I might be wrong due to vague arch understanding.

benjie: Good feedback; the reason I've gone with this approach is that "failed forever" tasks are expensive (they sit around forever making every scan for new jobs more expensive as the table is larger): they should be avoided where possible and periodically purged where not possible. This approach has been taken to improve the performance of the system, it's advised that jobs include their own logic for error tracking (or use the new events system #155) rather than scanning the database itself which will impact performance. Reporting systems can track the full job if they want (through the event system) and can see the attempts "drop down" to zero and track as appropriate.

PR #155, opened 2020-12-02
feat(events): add events system

### LINKED ISSUE #28: Add hooks/events
I am evaluation `worker` in conjunction with a table that holds a job "state" (created|retry|completed|failed|etc). For this it would be great to have global or queue events (or hooks?). At the moment I am not quite sure how to document a failed job without something like this. Possible events could be onError, onRetry, onFailed, onCompleted.

Maybe there are other use cases.

Something like:

```JS
async function main() {
  await run({
    connectionString: process.env.DATABASE_URL,
    concurrency: 1,
    pollInterval: 60000,
    taskList: tasks,
    events: {
      global: {
        onCompleted: async (job, result, helpers) => {} // Global events
      },
      queues : {
        myQueue: {
          onCompleted: async (job, result, helpers) => {} // Only for queue "myQueue"
        }
      }
    }
  });
}
```

Thoughts?

benjie:
## Description

There's lots of events that happen inside of Graphile Worker that library consumers might be interested in; this PR raises these events with an event system.

**IMPORTANT**: people registering event handlers **MUST** make sure that these handlers do **NOT** throw errors. We've wrapped the `emit` with a try/catch that swallows the error in the two most dangerous places, but consumers must be certain that errors are not thrown as this may cause unexpected states of the work queue to take place.

Events added so far:

- `pool:create` - When a worker pool is created
- `pool:listen:connecting` - When a worker pool attempts to connect to PG ready to issue a LISTEN statement
- `pool:listen:success` - When a worker pool starts listening for jobs via PG LISTEN
- `pool:listen:error` - When a worker pool faces an error on their PG LISTEN client
- `pool:release` - When a worker pool is released
- `pool:gracefulShutdown` - When a worker pool starts a graceful shutdown
- `pool:gracefulShutdown:error` - When a worker pool graceful shutdown throws an error
- `worker:create` - When a worker is created
- `worker:release` - When release of a worker is requested
- `worker:stop` - When a worker stops (normally after a release)
- `worker:getJob:error` - When a worker calls get_job but there are no available jobs
- `worker:getJob:empty` - When a worker calls get_job but there are no available jobs
- `worker:fatalError` - When a worker is created
- `job:start` - When a job is retrieved by get_job
- `job:success` - When a job completes successfully
- `job:error` - When a job throws an error
- `job:failed` - When a job fails permanently (emitted after job:error when appropriate)
- `gracefulShutdown` - When the runner is terminated by a signal
- `stop` - When the runner is stopped

Fixes #28 

## Performance impact

There must be an impact, but it seems too small to show up in benchmarks.

## Security impact

None known.

### DIFF
--- a/README.md
+++ b/README.md
@@ -409,6 +409,16 @@ export type WorkerEvents = TypedEventEmitter<{
   "worker:create": { worker: Worker; tasks: TaskList };
 
   /**
+   * When a worker release is requested
+   */
+  "worker:release": { worker: Worker };
+
+  /**
+   * When a worker stops (normally after a release)
+   */
+  "worker:stop": { worker: Worker; error?: any };
+
+  /**
    * When a worker calls get_job but there are no available jobs
    */
   "worker:getJob:error": { worker: Worker; error: any };
--- a/README.md
+++ b/README.md
@@ -262,16 +262,10 @@ the APIs for running jobs.
 ### `run(options: RunnerOptions): Promise<Runner>`
 
 Runs until either stopped by a signal event like `SIGINT` or by calling the
-`stop()` function on the Runner object `run()` resolves to.
+`stop()` method on the resolved object.
 
-The Runner object also contains a `addJob` method (see [`addJob`](#addjob)) that
-can be used to enqueue jobs:
-
-```js
-await runner.addJob("testTask", {
-  thisIsThePayload: true,
-});
-```
+The the resolved 'Runner' object has a number of helpers on it, see
+[Runner object](#runner-object) for more information.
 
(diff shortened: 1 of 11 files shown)


benjie:
This PR also re-orders some logic; I'm going to pull that out into a separate PR to keep this one cleaner.

shicholas:
this looks amazing, my only question above is how one would subscribe to the `EventEmitter`

benjie:
Great question @shicholas; does the above couple commits address this question sufficiently?

shicholas:
yes, thanks!

PR #161, opened 2020-12-09
feat(events): add more events

benjie:
## Description

Event just before get_job and just after fail_job/complete_job

## Performance impact

Negligible.

## Security impact

None.


mikeburgh:
Any thoughts to included the error object (if there was one) with the job:complete event ? I know job:error has it, but the idea of handling it after its been persisted to the DB appeals.

benjie:
PR welcome :+1:

PR #163, opened 2020-12-20
feat: regularly scheduled jobs (cron)

benjie:
## Description

Adds cron-like functionality to Graphile Worker, enabling you to schedule jobs on a recurring schedule. Read about it in detail in the included README changes.

This is intended to be used via a `crontab` file, an example of which is in the README, but you can also do it via the `crontab: string` library option, or if you need more advanced usage you can use the `parsedCrontabItems: ParsedCronItem[]` library option; to populate that you'd use `parseCrontab(crontab: string)` or `parseCronItems(items: CronItem[])`, for example:

```js
run({
  connectionString: "postgres://localhost/graphile_worker",
  parsedCronItems: parseCronItems([
    {
      task: "send_reminder",
      pattern: "10 10 * * *", // send at 10:10am every day
    },
  ]),
});
```

## Performance impact

If unused, should not impact main worker performance measurably. If used, could probably be more optimal, but it's pretty efficient.

## Security impact

No known issues.


Steven4294:
any eta on this getting merged?

src/crontab.ts
```diff
@@ -0,0 +1,396 @@
...
+export const parseCrontab = (crontab: string): Array<CronItem> => {
+  const lines = crontab.split(/\r?\n/);
+  const items: CronItem[] = [];
+  for (
+    let lineNumber = 1, numberOfLines = lines.length;
+    lineNumber <= numberOfLines;
+    lineNumber++
+  ) {
+    const line = lines[lineNumber - 1];
+    if (line.startsWith("#") || line.trim() === "") {
```
benjie, line 374: Need to think more about comments, particularly where `#` may occur in the rest of the line.

benjie: > Blank lines and leading spaces and tabs are ignored. Lines whose first non-space character is a pound-sign (#) are comments, and are ignored. Note that comments are not allowed on the same line as cron commands, since they will be taken to be part of the command. Similarly, comments are not allowed on the same line as environment variable settings.

Cannot do inline comments in cron :+1:

src/cron.ts
```diff
@@ -0,0 +1,530 @@
...
+ */
+export const runCron = (
+  options: RunnerOptions,
+  cronItems: CronItem[],
+  requirements: CronRequirements,
+): Cron => {
+  const { pgPool } = requirements;
+  const { logger, escapedWorkerSchema } = processSharedOptions(options);
+
+  // TODO: add events
```
benjie, line 276: TODO!

ben-pr-p (commented):
Although I agree with your design decisions here over https://github.com/davbeck/graphile-scheduler, I think @davbeck's Javascript land configuration API is easier to use (without included tasks or timezones):
```
run({
  connectionString: "postgres:///",
  schedules: [
    {
      name: "send_reminder",
      pattern: "0 10 * * 1-5",
    },
  ],
});
```

is easier for me than:
```
run({
  connectionString: "postgres:///",
  cronItems: [
    {
      task: "send_reminder",
      minutes: [10],
      hours: [10],
      dates: [...], // all dates must be included here, right?
      ...
    },
  ],
});
```

If I could delegate the cron parsing to the library but still have my task names typed in Typescript world that'd be the best of both worlds for me

benjie:
I don't really know how best to attack the remaining tests for this.

ncwp (dismissed):

ben-pr-p:
Thanks for working a parser into this! I'm sure it will save me at least some time if not a real bug 🥇

PR #192, opened 2021-06-01
docs(README): use @graphile/logger in logger example

mattiloh:
## Description

First things first: thanks @benjie for your great work on the graphile project 🙏 

After upgrading from `graphile-worker@0.9.0` to  `0.11.2` my custom logger setup failed with this message:
```
graphile_worker_1.Logger is not a constructor
TypeError: graphile_worker_1.Logger is not a constructor
```
I still imported Logger from `graphile-worker` as stated in the README, but as TS told me `'Logger' only refers to a type, but is being used as a value here.`

Installing `@graphile/logger` and importing the `Logger` from it fixed the issue for me. So I thought it might be helpful to update the docs.

## Checklist
n/a

### DIFF
--- a/README.md
+++ b/README.md
@@ -733,7 +733,8 @@ You may customise where log messages from `graphile-worker` (and your tasks) go
 by supplying a custom `Logger` instance using your own `logFactory`.
 
 ```js
-const { Logger, run } = require("graphile-worker");
+const { Logger } = require('@graphile/logger');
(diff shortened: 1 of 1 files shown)


benjie:
This is a bug; this should not have broken anything. We should be exporting a Logger instance.

benjie:
Thanks for the heads up on this :+1: Fixed in v0.11.3

PR #200, opened 2021-06-25
feat: allow using Node time instead of Postgres time (for tests)

hgl:
## Description

This PR adds a `{ now?: () => Date }` option that accepts an external time source. This can be helpful in tests where you want to artificially expire jobs.

For example:

If you pass `now: () => fakeClock.now` as an option, and then add a couple jobs that run in the future:

```
graphile_worker.add_job(...);
graphile_worker.add_job(...);
graphile_worker.add_job(...);
```

You can artificially tick time `fakeClock.tick("1 day")`, and Graphile Worker would find the expired jobs and run them. This is much easier to manage than having to manually update the `run_at` columns, which involves finding expired jobs and updating the column.

This PR is meant to start the conversation, so no tests or doc are added in the first commit. 

## Performance impact

unknown

## Security impact

unknown

## Checklist


ben-pr-p:
I see a use case for a fake now for `getJob`, which is to check that you have set your task up correctly in a test so that it will run in the future. In that case, the fake now should be passed all the way from `runTaskListOnce`, which is the utility I believe most people use for testing.

For us, querying the jobs table and asserting based on `run_at` has been enough, but I can imagine mocking now would be a good way to test a saga played out over time such as a welcome email drip, where a user gets an email right after signing up, 1 day after signing up, 3 days, etc.

I don't see the use case for adding a `now` parameter to `add_job`. `now()` here is only used as a fallback for `run_at`, so you should just be able to (and it's actively less confusing to) provide the `run_at` you need

hgl:
@ben-pr-p Thanks for the input.

> so you should just be able to (and it's actively less confusing to) provide the run_at you need

When `add_job` is run by tests, I agree that explicitly providing `run_at` is less confusing, but it can also be run by production code, which shouldn't be required to explicitly provide `run_at`. If you want to test such code with an external time source, I think it's less confusing that the default `run_at` time matches what the fake time is.

hgl:
Forgot to mention, currently cron jobs don't take external time source into consideration, in addition to an external `now`, they also need an external `setTimeout`. I guess the implementation should wait for the green light for the current commit first.

src/helpers.ts
```diff
@@ -29,7 +29,8 @@ export function makeAddJob(
           job_key => $6::text,
           priority => $7::int,
           flags => $8::text[],
-          job_key_mode => $9::text
+          job_key_mode => $9::text${!now ? "" : `,
+          now => $10::timestamptz`}
```
benjie, line 33: Let's make this less fluid:

```suggestion
          job_key_mode => $9,
          now => coalesce($10::timestamptz, now())
```

src/interfaces.ts
```diff
@@ -401,6 +401,11 @@ export interface SharedOptions {
    * An EventEmitter instance to which we'll emit events.
    */
   events?: WorkerEvents;
+
+  /**
+   * A time source for getting "now", can be helpful in tests to artificially expire jobs 
+   */
+  now?: () => Date
```
benjie, line 408: I don't think there's any need for this interface; instead use `jest-time-helpers` or similar and then you can just use `new Date()` as normal. I really don't want users to see `now` as an option in our configuration - it's purely a testing concern.

Instead let's say something like `useNodeTime` so it'll use the Node time instead of the Postgres time.

```suggestion

  /**
   * By default we use PostgreSQL's time source; in general this should be pretty close
   * (if not identical) to the time source your Node server is using, but in the case
   * that it isn't or you want to change it (e.g. in tests with fake date/time) you can
   * tell Worker to use Node's time source rather than Postgres' time source. (Default: 
   * false.)
   */
   useNodeTime?: boolean;
```

src/worker.ts
```diff
@@ -133,11 +134,12 @@ export function makeNewWorker(
           text:
             // TODO: breaking change; change this to more optimal:
             // `SELECT id, queue_name, task_identifier, payload FROM ...`,
-            `SELECT * FROM ${escapedWorkerSchema}.get_job($1, $2, forbidden_flags := $3::text[]); `,
+            `SELECT * FROM ${escapedWorkerSchema}.get_job($1, $2, forbidden_flags := $3::text[]${!now ? "": ", now :=> $4::timestamptz"}); `,
```
benjie, line 137: Let's always pass `now`:

```suggestion
            `SELECT * FROM ${escapedWorkerSchema}.get_job($1, $2, forbidden_flags := $3::text[]$, now :=> coalesce($4::timestamptz, now())); `,
```

src/worker.ts
```diff
@@ -133,11 +134,12 @@ export function makeNewWorker(
...
           values: [
             workerId,
             supportedTaskNames,
             flagsToSkip && flagsToSkip.length ? flagsToSkip : null,
+            ...(!now ? [] : [now().toISOString()])
```
benjie, line 142:
```suggestion
            useNodeTime ? new Date().toISOString() : null
```

benjie (changes requested):
Not a fan of the `now: () => Date` interface; should just use `new Date()` and have that change globally within the worker in tests. Let's instead express it as to whether we use PostgreSQL (`select now()`) or Node.js (`new Date()`) as the time source.

hgl:
I forgot how malleable JS APIs are. Relying on them being directly mockable is a good idea.

I'll try to make the changes within a week or two.

hgl:
Change implemented with cron support.

However, I don't feel comfortable for having introduced a `cron:loop` event which is made even more uglier by the fact that it requires `tickAsync`, but given the skewing nature of cron jobs, I'm not sure what design can be more elegant. Thoughts @benjie?

benjie:
Why do you need the `cron:loop` event; how do the existing `cron:*` events not serve your needs?

hgl:
Because this tick alone doesn't trigger a cron job's setTimeout

https://github.com/graphile/worker/blob/2b90b7e8b9605a8a3d8d7b8aba6a049059b5b139/__tests__/nodeTime.test.ts#L36-L37

Since the duration is usually around a minute.

And in order to trigger it, the ticking must happen after `setTimeout` is called. I don't think any existing `cron:*` events happen after `setTimeout`, and doesn't actually rely on it being triggered first.

hgl:
@benjie Sorry, I didn't understand how sinon's fake timer works well. I have pushed a commit to remove "cron:loop", and the test still passes.

The only caveat is that if the fake time doesn't start from the beginning of a minute, ticking must advance over the minute for the cron job to be added. Do you think it's something that should be documented or somehow addressed in code?

benjie (dismissed):
I've rewritten the tests - stripped out Sinon and replaced it with jest-time-helpers which we already use for the cron features (and I wrote...). I've also made the tests more granular, and added validation test that shows behaviour would be different without this option.

benjie:
> The only caveat is that if the fake time doesn't start from the beginning of a minute, ticking must advance over the minute for the cron job to be added. Do you think it's something that should be documented or somehow addressed in code?

This doesn't seem to be a concern with jest-time-helpers. Probably because I designed them to work specifically with Graphile Worker's cron support :grin:

benjie (dismissed):

hgl:
Thanks for the much comprehensive tests.

> This doesn't seem to be a concern with jest-time-helpers

Maybe there is some misunderstanding, I added a failing test to illustrate the issue. 

I think it's failed by

https://github.com/graphile/worker/blob/7a36910e702e70be6a03f5273dcfed58fccd027c/src/cron.ts#L330

benjie:
> The only caveat is that if the fake time doesn't start from the beginning of a minute, ticking must advance over the minute for the cron job to be added. Do you think it's something that should be documented or somehow addressed in code?

The cron should only fire for times hit after startup (when there is no backfilling), so we mustn’t set the test to start at the exact moment that a job should be scheduled (or even a fraction of a second earlier) because it will make the test unstable- sometimes it will match, sometimes not. This is expected behaviour: the ticking over of the minute is what triggers a cron.

>  I added a failing test to illustrate the issue.

Your test should start at 1:00:01am to ensure stability, and should advance by minutes, not seconds, since the `*/4` means every 4th minute. Also we require it to tick at least a millisecond past the minute in order to fire, this is to help absorb tiny amounts of clock skew.

hgl:
Somehow I thought `*/4` meant every 4th second, I must've been drunken.

> Your test should start at 1:00:01am to ensure stability, and should advance by minutes

This makes the test pass, and since it's not that different from the previous one, I've removed the commit.

benjie:
> I don't see the use case for adding a now parameter to add_job. now() here is only used as a fallback for run_at, so you should just be able to (and it's actively less confusing to) provide the run_at you need
> -- @ben-pr-p 

I agree with this. Adding a `now` argument is completely unnecessary; it'll only ever be used from JS and JS can just set the `run_at` instead. I've removed this change and have instead refactored the places that call `add_job` to pass the node time via run_at if no run_at is specified.

I was going to say "I hope the tests still pass" but amazingly they've already passed before I managed to finish writing this comment! Having fast tests is a delight :laughing:

benjie (approved):

Issue #198, opened 2021-06-11
Custom job locked interval


Ashniu123:
### Summary

Currently, if a job failed unexpectedly, blocks out the job for 4 hours (as per documentation "What if something goes wrong?"). Is there a possibility that we can set it to some custom value - say 1 hour?

### Additional context

Some critical tasks (short duration) run using graphile and would prefer shorter locked intervals. As a bonus, is there a way set locked_at per Runner, so that critical ones can be locked for short duration and long running ones for longer durations?

---
Thanks for building this amazing library based on SKIP LOCKED of pg. 🙌

benjie:
This is in fact an option to `get_job` in the database schema (`job_expiry`):

https://github.com/graphile/worker/blob/5eb792b1f31ff0b41e3591b34e25e87b73ce9c4c/__tests__/schema.sql#L223

but alas it is not yet configurable via the JavaScript API or CLI.

Specifically this is where it would need to be added to the JS:

https://github.com/graphile/worker/blob/5eb792b1f31ff0b41e3591b34e25e87b73ce9c4c/src/worker.ts#L131-L144

One reason I've not exposed this yet is because it's critical that this number is the same for all processes, and I wasn't happy with the risk of the figure coming out of sync and causing weird issues. A PR to add this feature would have to come with sufficient warnings in place to help ensure people don't shoot themselves in the foot accidentally - maybe in the form of documentation, maybe in the form of programmatic warnings, I'm honestly not sure.

Ashniu123:
Thanks for the quick reply @benjie really appreciate it! :raised_hands: 

Continuing on the reply...

Does this mean that even the "4 hour" interval is just a random value that you had thought of? There is no real reason for the interval to be either 4 hour or 2 hour or even 30 mins right?
The only consideration is that the `job` should finish before the interval, otherwise issues start popping up?

benjie:
Pretty much, yeah. It's designed to allow restarting crashed processes that were unable to report success/failure. It's meant to be very much a fallback that should virtually never be exercised (except when servers die without shutting down cleanly).

Ashniu123:
Hmmmm... To address my original question then, what do you think about the following:

1. Modify the `jobs` table to have an additional column `locked_interval` which would have a default value of 4 hours (like your implementation), but could be override with any other time interval value too (point 3).
2. Modify `get_job` to pick up job based on the row's `locked_at + locked_interval`
[...]

PR #209, opened 2021-08-18
allow custom job lock interval

Ashniu123:
## Description

Currently, each job has a default lock period of 4 hours. Some jobs may last longer than 4 hours and some may last only for a few minutes. With this feature, I aim to give the developer the power to control locked duration. This will allow graphile to execute jobs which last at most locked_duration, which is fixed to 4 now.

It is a highly risky thing to configure a task's lock period to be shorter than its execution time.

Related to issue #198 

## Performance impact

Unknown

## Security impact

Unknown

## Checklist

### DIFF
--- a/sql/000009.sql
+++ b/sql/000009.sql
@@ -1,0 +1,235 @@
+alter table :GRAPHILE_WORKER_SCHEMA.jobs add locked_duration interval;
+
+drop function :GRAPHILE_WORKER_SCHEMA.add_job(text, json, text, timestamptz, int, text, int, text[], text);
+create function :GRAPHILE_WORKER_SCHEMA.add_job(
+  identifier text,
+  payload json = null,
+  queue_name text = null,
+  run_at timestamptz = null,
+  max_attempts integer = null,
+  job_key text = null,
+  priority integer = null,
+  flags text[] = null,
+  job_key_mode text = 'replace',
+  locked_duration interval = null
+) returns :GRAPHILE_WORKER_SCHEMA.jobs as $$
+declare
+  v_job :GRAPHILE_WORKER_SCHEMA.jobs;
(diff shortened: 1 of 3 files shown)


gorbak25:
@Ashniu123 
Could you take a look at #210  ?

benjie:
Please run `yarn db:dump` and commit the updated dump, it makes the migration easier to read.

benjie:
So you think that rather than doing the lock duration on a per-queue or per-task-identifier basis that it should be on a per-job basis - basically it's up to whoever adds the job to ensure that they know how long it should take. I think that makes sense, but it's going to make the table bigger for fairly marginal gains, I fear. Do you have jobs that take more than an hour to complete?

Ashniu123:
> So you think that rather than doing the lock duration on a per-queue or per-task-identifier basis that it should be on a per-job basis - basically it's up to whoever adds the job to ensure that they know how long it should take. I think that makes sense, but it's going to make the table bigger for fairly marginal gains, I fear. Do you have jobs that take more than an hour to complete?

Yeah.

My take on having a custom job interval also comes from the viewpoint that if for any reason a process goes OOM which would cause jobs to be locked, a knowledgable developer should have the capability to tell graphile worker that the max run time of a job is 1 hour and to allow it to be picked up again asap.

Moreover, this PR would allow graphile worker to run jobs which last longer than 4 hours.

benjie:
Yeah; it's an interesting approach. I shall ponder it. I feel there's a better solution forming... leave it with me for a few weeks to brew.

Ashniu123:
Hey. Did think of a better solution yet? :monocle_face:

benjie:
Yes: https://github.com/graphile/worker/issues/222

PR #213, opened 2021-08-28
Fix cron not matching time properly for days of week

### LINKED ISSUE #212: Cron doesn't properly match days of the week
https://github.com/graphile/worker/blob/main/src/cron.ts#L586-L588

Here's the bug. The fix will be supplied soon.

wokalski:
Fixes #212

## Description

This is a breaking change in a sense, although it only fixes a feature that didn't work before.

### DIFF
--- a/__tests__/cron.test.ts
+++ b/__tests__/cron.test.ts
@@ -1,4 +1,5 @@
-import { run } from "../src";
+import { parseCronItem, run } from "../src";
+import { cronItemMatches, TimestampDigest } from "../src/cron";
 import {
   ESCAPED_GRAPHILE_WORKER_SCHEMA,
   EventMonitor,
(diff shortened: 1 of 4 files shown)


dawdus (approved):

benjie:
Awesome; thanks for spotting this typo! Took me a while to figure out what the specific issue was due to all the refactoring so I've split it out into #215 and #216. I've reverted the refactoring to `cronItemMatches` to make the fix clearer to those following along, but I don't think I'll reinstate that refactoring because it ever-so-slightly increases the CPU time required for those comparisons and since they run every minute I'm not willing to incur that cost (I know the cost is very very small).

benjie:
Fix released in graphile-worker@0.11.4 :raised_hands:

wokalski:
@benjie why do you think it takes more time CPU time in my refactor? I think it's exactly the same. `includes()` won't evaluate if dates or dows are exclusionary.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_AND

benjie:
```
  const dateIsExclusionary = cronItem.dates.length !== 31;
  const dowIsExclusionary = cronItem.dows.length !== 7;

  const dateMatches = !dateIsExclusionary || cronItem.dates.includes(date);
  const dowMatches = !dowIsExclusionary || cronItem.dows.includes(dow);
```

In your refactor the code above is evaluated before the

```
    cronItem.minutes.includes(min) &&
    cronItem.hours.includes(hour) &&
```

checks.

For most minutes and most jobs, the minutes check is the first thing that it's likely to stop on (since the minute won't match for most minutes) so in most cases this is the only check necessary; yours evaluates the length of two arrays and possibly does `.includes` on the arrays also (depending on the result) rather than just giving up on a single array lookup.

(I'm aware of short-circuiting logic; the current code depends on it.)

wokalski:
I find this kind of microoptimisations odd but I understand it.

benjie:
In a library that has performance as one of its main features, every little helps :smile:

PR #217, opened 2021-09-12
Add last execution time to _cron payload

### LINKED ISSUE #177: Find the last time a cron job was ran
## cron

I am setting a backfill period which makes sense for my application but additionally I'd like to set an alarm if the last execution of my cron was more than 3 days (my backfill period is 2 days). Is it possible yet?

wokalski:
## Description

Fixes #177 (although doesn't add the info to the event)

This PR adds the last execution and known since dates to the `_cron` payload. I have cron jobs that run every minute need to run back to back. So even if the server is down for some reason I still want to find the last execution date so that I properly report everything that happened since then.

Not all tests pass because I struggled to make `lastOnTimeExecution` contain the correct values. I wasn't aware that the CTE is more like a view rather than a table so when backfilling, last_exeuction gets changed as the query executes.

Any help would be appreciated here

## Performance impact

unknown

It adds a bit of complexity to the SQL query but also removes a bit of complexity from the JS code.

## Security impact

unknown

## Checklist

### DIFF
--- a/__tests__/cron.test.ts
+++ b/__tests__/cron.test.ts
@@ -98,7 +98,11 @@ test("backfills if identifier already registered (25h)", () =>
     await reset(pgPool, options);
     const now = Date.now();
     const expectedTime = now - (now % FOUR_HOURS);
-    await pgPool.query(
+    const {
+      rows: {
+        0: { last_execution: lastOnTimeExecution, known_since: knownSince },
+      },
+    } = await pgPool.query(
       `
         insert into ${ESCAPED_GRAPHILE_WORKER_SCHEMA}.known_crontabs (
           identifier,
@@ -110,6 +114,7 @@ test("backfills if identifier already registered (25h)", () =>
           NOW() - interval '14 days',
           NOW() - interval '25 hours'
         )
+        returning last_execution, known_since
       `,
     );
     const eventMonitor = new EventMonitor();
@@ -143,6 +148,20 @@ test("backfills if identifier already registered (25h)", () =>
       expect(jobs.length).toBeGreaterThanOrEqual(6);
       expect(jobs.length).toBeLessThanOrEqual(7);
       expect(jobs.every((j) => j.task_identifier === "do_it")).toBe(true);
+      expect(
+        jobs.every(
+          (j) =>
+            (j.payload as any)["_cron"]["knownSince"] ==
+            knownSince.toISOString(),
+        ),
+      ).toBe(true);
+      expect(
+        jobs.every(
+          (j) =>
+            (j.payload as any)["_cron"]["lastOnTimeExecution"] ==
+            lastOnTimeExecution.toISOString(),
+        ),
+      ).toBe(true);
     }
   }));
 
--- a/src/cron.ts
+++ b/src/cron.ts
@@ -73,22 +73,11 @@ function unsafeRoundToMinute(ts: Date, roundUp = false): Date {
   return ts;
 }
 
-function makeJobForItem(
-  item: ParsedCronItem,
-  ts: string,
-  backfilled = false,
-): CronJob {
+function makeJobForItem(item: ParsedCronItem): CronJob {
   return {
     task: item.task,
-    payload: {
-      ...item.payload,
-      _cron: {
-        ts,
-        backfilled,
-      },
-    },
+    payload: item.payload ?? {},
     queueName: item.options.queueName,
-    runAt: ts,
     maxAttempts: item.options.maxAttempts,
     priority: item.options.priority,
   };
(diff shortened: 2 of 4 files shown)


src/cron.ts
```diff
@@ -104,22 +93,39 @@ async function scheduleCronJobs(
...
+      with known_crontabs as (
+
+      ),
+      specs as (
         select
           index,
           (json->>'identifier')::text as identifier,
           ((json->'job')->>'task')::text as task,
-          ((json->'job')->'payload')::json as payload,
+          (((json->'job')->'payload')::jsonb || jsonb_build_object(
```
benjie, line 110: Casting between `json` and `jsonb` is expensive; if we can avoid it I'd rather we did.

Further, I'm pretty sure we already know the values from `known_crontabs` when scheduling the jobs, so requesting this data from the DB again should be unnecessary. We should be able to populate this `_cron` value entirely in JS.

src/cron.ts
```diff
@@ -104,22 +93,39 @@ async function scheduleCronJobs(
...
+                  known_crontabs.last_execution at time zone 'UTC', 'YYYY-MM-DD"T"HH24:MI:SS.MS"Z"'
+                ),
+                'knownSince', to_char(
+                  known_crontabs.known_since at time zone 'UTC', 'YYYY-MM-DD"T"HH24:MI:SS.MS"Z"'
+                )
+              )
+            )
+          )::json as payload,
           ((json->'job')->>'queueName')::text as queue_name,
-          ((json->'job')->>'runAt')::timestamptz as run_at,
```
benjie, line 119: `runAt` must be preserved.

src/interfaces.ts
```diff
@@ -476,11 +476,9 @@ export interface RunnerOptions extends WorkerPoolOptions {
 export interface CronJob {
   task: string;
   payload: {
-    _cron: { ts: string; backfilled?: boolean };
```
benjie, line 479: Why has this been removed?

src/interfaces.ts
```diff
@@ -476,11 +476,9 @@ export interface RunnerOptions extends WorkerPoolOptions {
...
     [key: string]: unknown;
   };
   queueName?: string;
-  runAt: string;
```
benjie, line 483: Why are we removing this? This definitely looks like a breaking change and I can't see the benefit?

benjie (changes requested):
Some initial feedback; I think this can be achieved with fewer breaking changes and less work for the database. I'm not sure that the database code needs to change at all, in fact.

wokalski:
Thanks for the feedback; one quick note: runAt is moved to the database layer fully. The rest of the feedback is valid. I'm pretty sure knownCrontabs are not available at the time when scheduling.

I'm also not sure whether I should put more stuff in JS or in the DB but I believe the less we pass the same arguments to multiple places the better. I.e. the runAt change seems like a nice simplification to me.

src/cron.ts
wokalski: I think we don't know these. If we did I'd use them for sure.

src/cron.ts
wokalski: It doesn't, it's supplied below.

src/interfaces.ts
wokalski: because the CronJob itself (as returned from the `make` function) doesn't contain it. In this implementation it's defined on the db layer.

src/interfaces.ts
wokalski: It's inserted in the DB query. No need to put the same information in multiple places IMO.

src/cron.ts
```diff
@@ -73,22 +73,11 @@ function unsafeRoundToMinute(ts: Date, roundUp = false): Date {
...
   return {
     task: item.task,
-    payload: {
-      ...item.payload,
-      _cron: {
-        ts,
-        backfilled,
-      },
-    },
+    payload: item.payload ?? {},
```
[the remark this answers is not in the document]
wokalski: this can be hit - if payload is undefined/null. Then on the DB layer we insert the `_cron` object inside the payload. If it's null it produces an array of two eleemnts. If it's an object, the `_cron` value is inserted into it.

__tests__/cron.test.ts
```diff
@@ -143,6 +148,20 @@ test("backfills if identifier already registered (25h)", () =>
...
+            knownSince.toISOString(),
+        ),
+      ).toBe(true);
+      expect(
+        jobs.every(
+          (j) =>
+            (j.payload as any)["_cron"]["lastOnTimeExecution"] ==
+            lastOnTimeExecution.toISOString(),
+        ),
+      ).toBe(true);
```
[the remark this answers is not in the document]
wokalski: great feedback.

benjie:
The DB is the bottleneck so we should avoid putting anything there that doesn't need to be. I'm planning to merge https://github.com/graphile/worker/pull/200 soon (honestly thought I had already merged it) so keeping runAt in JS will integrate much better with the external time source functionality.

src/cron.ts
benjie: Ah yes you're right; it's done as part of the CTE afterwards. Hmmm :thinking: I may need to think about this a bit more.

wokalski: Ok, please let me know what you think, I'll pause until then.

benjie: It's going to be a few days, I'm flat out at the moment :+1:

benjie: We _do_ know the last executed timestamp during backfilling - we get it during `registerAndBackfillItems`: https://github.com/graphile/worker/blob/5a09a37e94f49de0d531569f6ca00d5dadd816a5/src/cron.ts#L171

It's not exactly set up for this, in particular we don't create `backfillItemsAndDates` for `unknownIdentifiers` because we don't backfill them (they're unknown), so they're `knownSince: now, lastExecuted: never`. We could however add the relevant logic to enable this without needing much (if anything?) in the way of query changes.

However, we have to keep in mind that multiple workers are quite likely to be scheduling the same cron jobs at the same time, we do this on a "first call wins" basis via `on conflict ... where` in the `locks` CTE. That said, if we didn't hit that we can assume that someone else did so we can advance the `last_execution` timestamp. So I'm pretty sure we can factor this information into `makeJobForItem` calls and not have to change the `scheduleCronJobs` query at all.

That said... <see comment on PR as a whole>

benjie:
I don't think this is the right way to solve this problem, namely because it's misleading.

"last_execution" for a cron means when the cron fired and the job was _queued_.

"last_execution" for a job would be expected to mean when the job itself _executed successfully_.

These things only line up (roughly) when the job is executed on time and is successful on the first try. If the job is delayed or fails and retries then the information is misleading.

For https://pgRITA.com I have a script that handles the weekly emails, it deliberately tracks its results in the database and it uses this data as a "lock" that it controls. This means we can generate an email that represents to-the-second status of the users' data and the next email they receive will follow on from that very second, independent of when the jobs actually executed. Further if a job retries for long enough that another email has been sent in the interrim, the retrying job will automatically exit (successfully, without doing anything) because it knows its task has already been handled.

Basically what I'm saying is - upon further thought - I don't think I want this feature in Graphile Worker because it'll encourage people to build things in a less resilient way because it's _convenient_ to do so. Convenience can be a real footgun.

wokalski:
I am still planning to follow up with two-three smaller diffs.

PR #219, opened 2021-09-14
Add support for arbitrary functions in place of the cron pattern

### LINKED ISSUE #218: Arbitrary functions for cron patterns
### Feature description

Instead of just passing a cron pattern string we could also enable users to pass a function in the same place (i.e. "* * * * *" would be a shorthand for `cron("* * * * *")` where the function has to follow the `TimeDigest -> boolean` interface.

### Motivating example

One thing this solves in my specific use case are timezone-dependent patterns. I want to run a job in a given timezone, and make it work consistently even if the time offset changes due to daylight saving.

### Supporting development

- [x] am interested in building this feature myself

wokalski:
## Description

Fixes #218

Please keep in mind that this PR includes some changes. Cron expressions are not any different from any other function you can pass as the "schedule matcher". 

## Checklist

I haven't updated the documentation yet because I'm not sure if the implementation is satisfying.

### DIFF
--- a/__tests__/cron.test.ts
+++ b/__tests__/cron.test.ts
@@ -1,5 +1,4 @@
-import { parseCronItem, run, TimestampDigest } from "../src";
-import { cronItemMatches } from "../src/cronMatcher";
+import { run } from "../src";
 import {
   ESCAPED_GRAPHILE_WORKER_SCHEMA,
   EventMonitor,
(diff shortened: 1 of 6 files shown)


benjie:
This PR is quite hard to review due to the amount of refactoring; I've done some (opinionated) refactoring in https://github.com/graphile/worker/pull/220 that should mean you can get the diff in this PR down significantly. Please have a read of that PR description so you understand the reasoning behind the changes I made.

I'd really like this PR to contain only the changes necessary for the feature and avoid refactoring where possible - if this means a couple more precursor PRs I'm cool with that - I'd rather review more smaller PRs than one large PR.

While you're doing the rebase, please could you rename `CronSchedule` to `CronMatcher` - the latter sounds a lot more like a function than the former to me at least.

wokalski:
Makes sense. I will try to do a couple of small PRs.

wokalski:
The timezone change from this weekend is just another reason for me to add this 😄 ... I'll get back to it before the end of the year. It's quite clear how I can split this into 2-3 smaller PRs.

wokalski:
@benjie I rebased this on top of #250 as a single-PR alternative.

wokalski:
This PR is net 26 lines and adds a very valuable feature. I hope we can merge it this time.

benjie (commented):
Let's get #250 merged first, but I don't think I have any major qualms with this; the main issue would be that I want to maintain the performance of cron, but I don't believe this causes any issue for this (and, in fact, it potentially gives us the ability to make cron _faster_!)

PR #223, opened 2021-09-25
Release worker & cron before release PG pool/connections when stopping runner

countcain:
## Description

Currently, when stopping a runner, the releases of worker pool, cron, pg pool happens at the same time. I think it is executed [here](https://github.com/graphile/worker/blob/main/src/lib.ts#L147-L149).

However, during my test, it turns out if there is an active job in the worker pool when `stop` is called. The release of the worker pool will wait for that active job to be processed. And at the end of job processing, it writes job status back to PG. If successful, it removes the job record, if failed, it updates the error message. But pg pool is ended at that moment.

Here is a code snippet I use:

```javascript
const { run, Logger, makeWorkerUtils, quickAddJob } = require("graphile-worker");
const EventEmitter = require('events')

const pgURL = "postgres://postgres:my_password@dell.home:5432/dev"
const maxAttempts = 3

async function main() {

  const workerUtils = await makeWorkerUtils({
    connectionString: pgURL,
  })
  await workerUtils.migrate()

  const events = new EventEmitter()
  events.on('job:start', ({worker, job})=>{
    console.log('start', worker.workerId, job.payload.name, Date.now() - job.created_at)
  })
  events.on('job:failed', async ({job}) => {
    console.log('permanently failed')
    // remove permanently failed job record in case the number of records keeps growing which will slow down worker performance on retrieving jobs
    await workerUtils.completeJobs([job.id])
  })

  const runner = await run({
    connectionString: pgURL,
    concurrency: 5,
    logger: new Logger((scope)=> (level, message , meta)=>{
      // console.log(scope, level, message, meta)
    }),
    noHandleSignals: true,
    pollInterval: 1000,
    events,
    taskList: {
      hello: async (payload, {job}) => {
        try {
          await new Promise((resolve, reject)=>{
            console.log('processing', job.id)
            setTimeout(()=>{
              if (payload.name === 'success' || (job.attempts === job.max_attempts && payload.name !== 'hey1')) {
                return resolve()
              }
              reject(new Error('timeout'))
            }, 10000)
          })
          await new Promise(resolve => {
            setTimeout(()=>{
              // just mocking
              console.log('saved success jobs to DB')
              resolve()
            }, 100)
          })
        } catch(e) {
          if (job.attempts === job.max_attempts) {
            await new Promise(resolve => {
              setTimeout(()=>{
                // just mocking
                console.log('saved permanent job failures to DB')
                resolve()
              }, 100)
            })
          }
          throw e
        }
      },
    },
  });

  const add = n => quickAddJob({
      connectionString: pgURL,
    }, 'hello',
    {
      name: `hey${n}`
    },
    {
      queueName: '1',
      maxAttempts
    }
  )

  process.on('SIGTERM', async () => {
    try {
      await runner.stop()
    } catch (e) {
      console.log('stop failed', e)
    }
    await runner.promise
    process.exit(0)
  })

  await runner.addJob('hello', {
    name: 'success'
  }, {
    queueName: '1',
    maxAttempts,
  })

  await add(1)
  await add(2)
  await add(3)
}

console.log(process.pid)
main().catch((err) => {
  console.error(err);
  process.exit(1);
});
```

Here is the screenshot of running this code snippet

Screen Shot 2021-09-24 at 11 24 12 PM

Here are job entries in DB

Screen Shot 2021-09-24 at 11 59 01 PM

You can see that due to the pg pool being closed, the worker cannot update the job record, it remains locked.

## Checklist

### DIFF
--- a/src/runner.ts
+++ b/src/runner.ts
@@ -87,10 +87,8 @@ export const run = async (
       (await getParsedCronItemsFromOptions(options, releasers));
 
     const cron = runCron(options, parsedCronItems, { pgPool, events });
-    releasers.push(() => cron.release());
 
     const workerPool = runTaskList(options, taskList, pgPool);
-    releasers.push(() => workerPool.release());
 
     let running = true;
     return {
@@ -98,6 +96,7 @@ export const run = async (
         if (running) {
           running = false;
           events.emit("stop", {});
+          await Promise.all([cron.release(), workerPool.release()])
           await release();
         } else {
           throw new Error("Runner is already stopped");


benjie:
Good catch!

I think I might go about this a different way - e.g. release things in the reverse order that they were added to releasers. Currently it's done in parallel but as you point out that's clearly wrong. Or maybe they need to register some kind of "priority". Or maybe we _only_ need to release the pool last and everything else can be done in parallel...

Let's start with a smaller change that just releases things in reverse order (waiting for completion before moving to the next) - would you care to make the changes? If any of the releases fail (throw an error) please continue on to the next release regardless; you can then re-raise whatever the first error was at the end of the release process.

countcain:
> Before I merge; can you confirm that this change fixes the issue you were facing?

I've verified the fix and it does solve the issue. Looking forward to the merge and a new release. Thank you @benjie

PR #242, opened 2021-12-07
fix: handle errors when releasing worker

countcain:
## Description

First of all, thank you for your library. I discovered a bug today.

Screen Shot 2021-12-07 at 6 05 04 PM

The callback passed to `pool.connect(callback)` [in this case](https://github.com/brianc/node-postgres/blob/master/packages/pg-pool/index.js#L166) will only be called with an error. Therefore, `releaseClient` can be undefined.

## Performance impact

n/a

## Security impact

n/a

## Checklist

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -197,7 +197,7 @@ export function runTaskList(
   const listenForChanges = (
     err: Error | undefined,
     client: PoolClient,
-    releaseClient: () => void,
+    releaseClient: () => void = () => {},
   ) => {
     if (!active) {
       // We were released, release this new client and abort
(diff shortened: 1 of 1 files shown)


benjie (changes requested):
Good catch; please remove the default parameter and instead change line 204 to `releaseClient?.();` - the issue here is that the `!active` handling is triggered before the `err` is handled. (Technically the types are wrong and we should define it as an overloaded function, but that seems a lot of effort.)

countcain:
@benjie I see. I've updated L204.

benjie (approved):
Great spot 👍

countcain:
thank you! It would be very nice if you can merge this and release a patch version.

benjie:
Oh gosh sorry, this completely got buried :grimacing: Thanks for the fix - it's out now in 0.12.2.

PR #250, opened 2022-02-20
Internal refactor of cron

wokalski:
## Description

This was performed to prepare for proper implementation of #218. No new functionality was
implemented as part of this commit.

## Performance impact

should be none

## Security impact

unknown

## Checklist

- [x] I've added tests for the new feature, and `yarn test` passes.`


src/cronMatcher.ts
```diff
@@ -1,16 +1,26 @@
...
+import { TimestampDigest } from "./interfaces";
+
+// Crontab ranges from the minute, hour, day of month, month and day of week parts of the crontab line
+interface CrontabRanges {
+  minutes: number[];
+  hours: number[];
+  dates: number[];
+  months: number[];
+  dows: number[];
+}
```
benjie, line 16: Please do the following:

1. Rename this from `CrontabRanges` to `ParsedCronMatch`
2. Move this to interfaces just above `ParsedCronItem`
3. Add `extends ParsedCronMatch` to the interface definition of `ParsedCronItem` 
4. Move the comments and definitions for these 5 items from `ParsedCronItem` to `ParsedCronMatch`

src/cronMatcher.ts
```diff
@@ -1,16 +1,26 @@
...
 
 /**
  * Returns true if the cronItem should fire for the given timestamp digest,
  * false otherwise.
  */
 export function cronItemMatches(
-  cronItem: ParsedCronItem,
+  cronItem: CrontabRanges,
```
benjie, line 23: Note to self: this is *not* a breaking change.

src/cronMatcher.ts
```diff
@@ -170,3 +180,11 @@ export function parseCrontabRanges(matches: string[], source: string) {
   );
   return { minutes, hours, dates, months, dows };
 }
+
+export const parseCronString = (pattern: string, source: string = "") => {
```
benjie, line 184: Please add an explicit return type.

benjie: Also, please rename to `parseCronRangeString` since it's only responsible for parsing the "range" part of a cron string :+1:

benjie (changes requested):
Solid changes; just a bit of painting the bike shed and we're good to go :+1:

src/cronMatcher.ts
wokalski: good suggestion!

wokalski:
hey @benjie can we get this merged? I'll sort out the CI but still needs an approval

src/cronMatcher.ts
```diff
@@ -170,3 +174,14 @@ export function parseCrontabRanges(matches: string[], source: string) {
   );
   return { minutes, hours, dates, months, dows };
 }
+
+export const parseCronRangeString = (
+  pattern: string,
+  source: string = "",
```
benjie, line 180: We should only add `in ${source}` if there is a source; so either remove the `= ""` here or handle the `source` missing in the error message and make sure `parseCrontabRanges` is also happy with no source.

```suggestion
  source: string,
```

benjie (changes requested):
Sorry! I've been super busy on the PostGraphile V5 repo and am really behind on GitHub notifications right now. This is looking good 👍

Issue #208, opened 2021-08-13
Add ability to override the schema in which extensions are installed


jcgsville:
### Feature description

I'd like the ability to control via some mechanism which schema the pgcrypto schema is attempted to be installed into. Today, it is only installed into `public`.

https://github.com/graphile/worker/blob/e3176eab42ada8f4f3718192bada776c22946583/src/migrate.ts

It makes no difference to me if this is an environment variable or cli flag/library option.

### Motivating example

For my graphile-based project, I went down the path of removing the public schema entirely from my DB. I installed all extensions into an `extensions` schema and put only this schema on the search path. While it's not necessarily typical, some quick googling suggests that I'm not the only one who has taken this approach, so it seems reasonable that graphile worker should support this.

### Breaking changes

This should not involve breaking changes, the default schema would remain `public`

### Supporting development

I [tick all that apply]:

- [x] am interested in building this feature myself
(I'm not chomping at the bit to take this on, but I'm willing to do it to get off my fork 😄 )
- [x] am interested in collaborating on building this feature
- [x] am willing to help testing this feature before it's released
- [ ] am willing to write a test-driven test suite for this feature (before it exists)
- [ ] am a [Graphile sponsor](https://www.graphile.org/sponsor/) ❤️
- [ ] have an active [support or consultancy contract](https://www.graphile.org/support/) with Graphile

jcgsville:
I did a little bit of digging here, and it looks like this change would require not just updating migrate.ts, but also the migration SQL files. Is that something we're open to doing? Or are migrations append-only?

In reading through migrate.ts, it seems like it should be fine and wouldn't affect anyone who's run the migration already, but there could be something I'm missing there.

benjie:
We already use `:GRAPHILE_WORKER_SCHEMA` in the migrations to changing `public.gen_random_uuid` to `:PG_CRYPTO_SCHEMA.gen_random_uuid` seems reasonable. You can do that to pre-existing migrations, they're not signed or anything like that :eyes: :grin:

jcgsville:
Closing in favor of #251

PR #251, opened 2022-02-21
feat: Don't install pgcrypto

### LINKED ISSUE #249: pgcrypto not necessary in PostgreSQL 13+?
Correct me if I'm wrong, but pgcrypto only appears to be installed in order to use `gen_random_uuid()`. That function [is available without the extension](https://www.postgresql.org/docs/current/functions-uuid.html) in Postgres 13+. Is it feasible to make the installation of pgcrypto conditional on version so it can be avoided if unnecessary?

noinkling:
## Description

Fixes #249

Some notes:

- I couldn't simply add a new migration file because migrations were failing (PG13/no pgcrypto) in tests after removing the `create extension` statement from the migration runner. Instead I had to modify 000001.sql directly. ~~This means there would be no easy migration strategy for users, they would need to drop the schema and have it recreated. I'm not sure if this is acceptable.~~

  Edit: Realized I can create the new migration anyway, users will be able to update normally.

- ~~There are a couple more instances of `public.gen_random_uuid()` in 000002.sql, but since they're in function bodies they don't cause the migration process to fail, and those functions get overwritten later which removes those calls anyway. I thought it better to just leave that file as is.~~

- ~~The schema.sql dump was regenerated with PG13. If it were regenerated with an earlier version that uses pgcrypto I assume the function call would be prefixed with `public.` again (as pg_dump usually does, since it also clears the search path). I see the file's only purpose is informational, so maybe that doesn't matter.~~

## Performance impact

unknown

## Security impact

unknown

## Checklist


benjie:
Please can you modify all the migrations to remove `public.`, and add to the new migration `create or replace function` statements for all functions into their current (non-`public.`) state. I want to know that at the end of this all the Graphile Worker schemas will be identical.

As you note, the `schema.sql` is primarily to help me do code review :wink:

noinkling:
> Please can you modify all the migrations to remove `public.`

Done.

> and add to the new migration `create or replace function` statements for all functions into their current (non-`public.`) state.

There are four references total to `gen_random_uuid()` in the migrations. Three of those appear in `.add_job()` which is replaced (multiple times) in later migrations, removing those references already. The definition for `.add_job()` would just be the same as the most recent one in 000007.sql

The only remaining reference once all the migrations have run is the default value for the `queue_name` column on the `jobs` table, which you can confirm by looking at `schema.sql`. That's why it's the only one I've changed in the new migration.

----

As for the CI failures, I'm pretty sure those are just because pgcrypto isn't being installed automatically anymore (it probably also breaks the Docker workflow if you don't install it manually). Tests pass locally because I added a `setup-prereqs.sql` script to the `yarn test` command, which conditionally installs it. However the CI just runs `jest`.

So to fix it I'm thinking I make use of Jest's [`globalSetup` option](https://jestjs.io/docs/configuration#globalsetup-string) instead. Actually I was considering doing this from the start, but I wanted to see if I could get away with the SQL script 😛

__tests__/helpers.ts
```diff
@@ -24,10 +24,6 @@ export {
   WEEK,
 } from "jest-time-helpers";
 
-// Sometimes CI's clock can get interrupted (it is shared infra!) so this
-// extends the default timeout just in case.
-jest.setTimeout(15000);
```
noinkling, line 29: This had to be removed because the `jest` global isn't available in the global setup environment, which utilizes this module. It doesn't seem like the ideal place for it anyway. Instead I changed the `testTimeout` option in the config file.

noinkling:
I changed it to use Jest's global setup, should be good for another CI run.

benjie (commented):
This looks good to me, why is it a draft?

noinkling:
I was waiting to see the CI results, and I haven't yet added the change to the release notes and wasn't sure if it needed to be tested explicitly.

The database_updated CI job is still failing because it's running on PG11 and I didn't update the dump_db script to install pgcrypto. I'll admit, when I ran that locally to regenerate schema.sql I was using PG13, rather than PG11 as instructed in the readme. I'll fix it up when I get the chance.

benjie:
Fantastic - best of luck :+1:

jcgsville:
Does this PR make #208 moot if you're willing to use pg 13+?

benjie:
It may well do, yes.

By the way CI is failing because the schema.sql needs to be updated.

noinkling:
Ready for another CI run, hopefully all passing this time.

Since I generated schema.sql with PG11 this time, it went back to how it was originally. Just be aware that generating it with PG13+ will give a slightly different result.

noinkling:
@benjie I have a question that I probably should have asked earlier: is `gen_random_uuid()` needed at all currently?

The only reference to it is as the default value for the `queue_name` column on the `jobs` table, and as far as I can tell the current APIs for adding/updating a job don't have any way to trigger it, they insert null as the default/fallback instead. I believe a user would need to insert/update a job manually to make use of it. I just assumed that it was still being useful somehow, could it simply be a leftover that was never cleaned up? e.g. from #77

benjie:
(Sorry for the delay, I'm snowed under right now :sweat_smile:)

You're absolutely right - we should drop the default and then drop the need for it entirely - edit it out of previous migrations, make a new migration that updates existing schemas to not reference it, and then we're done I hope :+1:

benjie:
Solved by #260

PR #260, opened 2022-04-24
feat(deps): remove dependency on gen_random_uuid()/pgcrypto

### LINKED ISSUE #249: pgcrypto not necessary in PostgreSQL 13+?

noinkling:
## Description

Alternative to/supersedes #251, as per [comment](https://github.com/graphile/worker/pull/251#issuecomment-1106608649). Fixes #249.

I only edited the existing migrations the minimum amount to prevent them from failing. It's probably possible to remove large parts of the old cruft, but as you are no doubt aware, the idea of migrations is not to edit them later (ideally), and you would have to be more careful not to break users upgrading from an old version, so I didn't attempt that.

## Performance impact

unknown

## Security impact

unknown

## Checklist

- [x] ~~I've added tests for the new feature, and~~ `yarn test` passes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,7 +210,7 @@ PostgreSQL 10+\* and Node 10+\*.
 
 _Note: `graphile-worker` versions 0.12.2 and earlier installed the `pgcrypto`
 extension into the public schema of your database (if it wasn't already
-installed). As of `graphile-worker` version _\<pending\>_ we no longer use
+installed). As of `graphile-worker` version 0.13.0 we no longer use
 `pgcrypto`. Existing users may want to uninstall it - see the
 [release notes](RELEASE_NOTES.md#pending) for instructions._
 
--- a/README.md
+++ b/README.md
@@ -208,11 +208,11 @@ http://discord.gg/graphile
 
 PostgreSQL 10+\* and Node 10+\*.
 
-If your database doesn't already include the `pgcrypto` extension we'll
-automatically install it into the public schema for you. If the extension is
-installed in a different schema (unlikely) you may face issues. Making alias
-functions in the public schema, should solve this issue (see issue
-[#43](https://github.com/graphile/worker/issues/43) for an example).
+_Note: `graphile-worker` versions 0.12.2 and earlier installed the `pgcrypto`
+extension into the public schema of your database (if it wasn't already
+installed). As of `graphile-worker` version _\<pending\>_ we no longer use
+`pgcrypto`. Existing users may want to uninstall it - see the
+[release notes](RELEASE_NOTES.md#pending) for instructions._
 
 \* Might work with older versions, but has not been tested.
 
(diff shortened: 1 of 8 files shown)


sql/000001.sql
```diff
@@ -79,7 +79,7 @@ create trigger _900_notify_worker after insert on :GRAPHILE_WORKER_SCHEMA.jobs f
 create function :GRAPHILE_WORKER_SCHEMA.add_job(
   identifier text,
   payload json = '{}',
-  queue_name text = public.gen_random_uuid()::text,
+  queue_name text = 'PLACEHOLDER',
```
noinkling, line 82: This feels weird, but technically at this point in time the column is non-nullable so I used a string that makes it clear that it isn't actually going to be used.

benjie: Use `null /* was gen_random_uuid(), but later removed dependency */`

README.md
```diff
@@ -208,11 +208,11 @@ http://discord.gg/graphile
...
 PostgreSQL 10+\* and Node 10+\*.
 
-If your database doesn't already include the `pgcrypto` extension we'll
-automatically install it into the public schema for you. If the extension is
-installed in a different schema (unlikely) you may face issues. Making alias
-functions in the public schema, should solve this issue (see issue
-[#43](https://github.com/graphile/worker/issues/43) for an example).
+_Note: `graphile-worker` versions 0.12.2 and earlier installed the `pgcrypto`
+extension into the public schema of your database (if it wasn't already
+installed). As of `graphile-worker` version _\<pending\>_ we no longer use
```
benjie, line 213:
```suggestion
installed). As of `graphile-worker` version 0.13.0 we no longer use
```

RELEASE_NOTES.md
```diff
@@ -1,5 +1,12 @@
 # Release notes
 
+### Pending
+
+- Remove dependency on `pgcrypto` database extension
+  - If you have a pre-existing installation and wish to uninstall `pgcrypto` you
+    will need to do so manually. This can be done by running
+    `DROP EXTENSION pgcrypto;` _after_ updating to the latest schema.
```
benjie, line 8:
```suggestion
    `DROP EXTENSION pgcrypto;` _after_ updating to the latest schema.
- jobs.queue_name no longer has a default (this is only relevant to people inserting to the table directly, which is not recommended - use the add_job helper)
```

benjie (changes requested):
Looking good!

benjie (approved):
Excellent; this will make it even easier for people to adopt Graphile Worker - thanks! :raised_hands:

benjie:
Released in v0.13.0 :raised_hands:

PR #267, opened 2022-06-08
docs(readme): Clarify `maxAttempts`

### LINKED ISSUE #266: Does `maxAttempts` include initial attempt?
Graphile works great for us at PostHog, but we'd actually like to disable the built-in retries feature, so that users of our system can define retries themselves. However, having read the docs, I'm not quite sure what `maxAttempts` means. Concretely, does `maxAttempts: 1` mean that the job will be executed at most once (just the initial attempt), or at most twice (the initial attempt + up to 1 retry)?

Twixes:
## Description

Tiny `README` enhancement to resolve #266.

### DIFF
--- a/README.md
+++ b/README.md
@@ -691,7 +691,8 @@ export interface TaskSpec {
   priority?: number;
 
   /**
-   * How many attempts should this task get? The minimum is 1, in which case the task will only be attempted once and won't be retried. (Default: 25)
+   * How many attempts should this task get? The minimum is 1, in which case the
+   * task will only be attempted once and won't be retried. (Default: 25)
    */
   maxAttempts?: number;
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -691,7 +691,7 @@ export interface TaskSpec {
   priority?: number;
 
   /**
-   * How many retries should this task get? (Default: 25)
+   * How many attempts should this task get? The minimum is 1, in which case the task will only be attempted once and won't be retried. (Default: 25)
```
benjie, line 694:
```suggestion
   * How many attempts should this task get? The minimum is 1, in which case the
   * task will only be attempted once and won't be retried. (Default: 25)
```

benjie (changes requested):
Perfect wording!

benjie (approved):

PR #271, opened 2022-06-24
chore: internal refactor of cron

benjie:
This is #250 but with the two additional fixes applied; unfortunately I was unable to push to the source branch so I've raised a separate PR. Please see #250 for discussion/etc.

Fixes #250


benjie:
Thanks @wokalski!

wokalski:
Sorry i didnt get back to it @benjie. Too little time in a day!

PR #272, opened 2022-06-24
feat(cron): allow arbitrary functions in place of the cron pattern

benjie:
This is a re-implementation of #219 on top of the recently merged #271.

Credit to @wokalski for the idea and the initial implementation.

This is a breaking change because `pattern` has been renamed to `match`. (Our next release of worker will be a breaking change anyway, so may as well bundle it in!) No other external interface has changed (other than a few new internal types being available due to `export *`...)

Fixes #219


wokalski:
we are very happy for this feature!

PR #304, opened 2022-11-17
docs: fix example trigger function

igor47:
from https://www.postgresql.org/docs/current/plpgsql-trigger.html:
> In the case of a before-trigger on DELETE, the returned value has no direct effect, but it has to be nonnull to allow the trigger action to proceed. Note that NEW is null in DELETE triggers, so returning that is usually not sensible. The usual idiom in DELETE triggers is to return OLD.


benjie:
> In the case of a before-trigger on DELETE,

These are all after triggers, not before triggers.

PR #309, opened 2022-12-22
feat: add Postgres version check

maxkorp:
## Description

Adds a check which throws an error if postgres 11 or lower is detected. This runs before migrating, because migration 11 throws a weird syntax error on pg10/11. 

As a small additional fix from testing, the Dockerfile.dev file was requiring node 12, which throws an error with yarn version constraints, so I upgraded that to node:12-alpine to align with the rest of the project. 

Not sure this qualifies as a breaking change, as the requirement for pg12 already existed, this just causes a clearer error. 

Not really sure how to add a test for this, I just modified the docker-compose.yml file to pull postgres 10 and 11 instead of 12 (and 13/14). Happy to add something like that, just not sure how you'd want to structure that. 

## Performance impact

None that I can find, theoretically a minimal startup time cost. 

## Security impact

None, should be low since it's just a query of `show server_version`

## Checklist


src/migrate.ts
```diff
@@ -6,13 +6,28 @@ import { processSharedOptions } from "./lib";
...
       id int primary key,
       ts timestamptz default now() not null
     );
+    select current_setting('server_version');
   `);
+
+  checkPostgrsVersion(row.server_version);
+}
+
+function checkPostgrsVersion(versionString: string) {
```
purge, line 23: (sp)

maxkorp: Yeah that's fixed >.<

src/migrate.ts
```diff
@@ -56,18 +78,20 @@ export async function migrate(
     const {
       rows: [row],
     } = await client.query(
-      `select id from ${escapedWorkerSchema}.migrations order by id desc limit 1;`,
+      `select current_setting('server_version_num') as server_version_num,
+      (select id from ${escapedWorkerSchema}.migrations order by id desc limit 1);`,
```
benjie, line 82:
```suggestion
      (select id from ${escapedWorkerSchema}.migrations order by id desc limit 1) as id;`,
```

benjie (changes requested):

benjie (approved):

PR #347, opened 2023-10-03
chore: fixes to previous PR

benjie:
Follow up to #345 with a couple of fixes (since I could not push to that branch).


timelf123:
Thanks for this! Came back to make a few tweaks and you had beaten me to it

benjie:
Yeah; it should have been obvious to me that `\` won't work in `` String.raw`...` `` calls... That's kind of the point! Is it too late to blame jetlag? :grin:

timelf123:
Computers are hard! We are telling sand what to do and sometimes that gets lost in translation :P Thanks again

PR #404, opened 2023-11-23
website: Docs tweaks to AddJob and TypeScript pages

jemgillam:
## Description

Some documentation tweaks:

- `addJob()` example now mirrors the example on the "Add jobs through SQL" page
-  changed `payload` to `unknown` type with an explanation, on the same `AddJob()` page
- Reordered the TypeScript page to put the recommended route at the top
- Added an Amazon SES example to the TypeScript page

## Performance impact

Docs only

## Security impact

Docs only

## Checklist

- ~~[ ] I've added tests for the new feature, and `yarn test` passes.~~
- ~~[ ] I have added this feature to 'Pending' in the `RELEASE_NOTES.md` file (if one exists).~~
- ~~[ ] If this is a breaking change I've explained why.~~

### DIFF
--- a/website/docs/typescript.md
+++ b/website/docs/typescript.md
@@ -74,8 +74,8 @@ export const send_email: Task = async function (payload) {
   const { to, subject, body } = payload;
   await ses.sendEmail({
     Destination: {
-      ToAddress: [to],
-      FromAddress: ["no-reply@example.com"],
+      ToAddresses: [to],
+      FromAddresses: ["no-reply@example.com"],
     },
     Message: {
       Subject: {
@@ -123,9 +123,9 @@ export const send_email: Task = async function (payload) {
 + const { to, subject, body, from } = payload;
   await ses.sendEmail({
     Destination: {
-      ToAddress: [to],
--     FromAddress: ["no-reply@example.com"],
-+     FromAddress: [from ?? "no-reply@example.com"],
+      ToAddresses: [to],
+-     FromAddresses: ["no-reply@example.com"],
++     FromAddresses: [from ?? "no-reply@example.com"],
     },
     Message: {
       Subject: {
@@ -222,8 +222,8 @@ export const send_email: Task<"send_email"> = async function (payload) {
   const { to, subject, body } = payload;
   await ses.sendEmail({
     Destination: {
-      ToAddress: [to],
-      FromAddress: ["no-reply@example.com"],
+      ToAddresses: [to],
+      FromAddresses: ["no-reply@example.com"],
     },
     Message: {
       Subject: {
@@ -267,9 +267,9 @@ export const send_email: Task<"send_email"> = async function (payload) {
 + const { to, subject, body, from } = payload;
   await ses.sendEmail({
     Destination: {
-      ToAddress: [to],
--      FromAddress: ["no-reply@example.com"],
-+      FromAddress: [from ?? "no-reply@example.com"],
+      ToAddresses: [to],
+-      FromAddresses: ["no-reply@example.com"],
++      FromAddresses: [from ?? "no-reply@example.com"],
     },
     Message: {
       Subject: {
--- a/website/docs/typescript.md
+++ b/website/docs/typescript.md
@@ -10,38 +10,17 @@ ensure the `payload` conforms to what you expect. It can be convenient to
 declare the payload types up front to avoid this `unknown`, but doing so might
 be unsafe &mdash; please be sure to read the caveats below.
 
-## Using type guards
+## Using type assertion functions
 
 To ensure your system is as safe as possible (and guard against old jobs, or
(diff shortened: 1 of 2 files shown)


website/docs/typescript.md
```diff
@@ -8,12 +8,146 @@ typed as `unknown` since they may have been populated by out of date code, or
...
+  if (typeof payload.body !== "string") throw new Error("invalid");
+}
+
+export const send_email: Task = async function (payload) {
+  assertPayload(payload);
+  const { to, subject, body } = payload;
+  await ses.sendEmail({
+    Destination: {
+      ToAddress: [to],
+      FromAddress: ["no-reply@example.com"],
```
benjie, line 78:
```suggestion
      ToAddresses: [to],
      FromAddresses: ["no-reply@example.com"],
```

website/docs/typescript.md
```diff
@@ -8,12 +8,146 @@ typed as `unknown` since they may have been populated by out of date code, or
...
+        },
+      },
+    },
+  });
+};
+```
+
+If now we introduce a new functionality to set the from address, the changes we
+make have to take into account that older jobs may not have the from address
+set, like so:
```
benjie, line 98:
```suggestion
If now we introduce a new functionality to set the `from` address, we have to
take into account that older jobs will not have the `from` address set. We
should adjust our code like so:
```

website/docs/typescript.md
```diff
@@ -56,47 +190,99 @@ created in the database directly via the `graphile_worker.add_job()` or
...
 
-const task: Task = async (payload) => {
-  assertMyPayload(payload);
-  console.log(payload.username);
+export const send_email: Task<"send_email"> = async function (payload) {
+  const { to, subject, body } = payload;
+  await ses.sendEmail({
+    Destination: {
+      ToAddress: [to],
+      FromAddress: ["no-reply@example.com"],
```
benjie, line 226:
```suggestion
      ToAddresses: [to],
      FromAddresses: ["no-reply@example.com"],
```

website/docs/typescript.md
```diff
@@ -56,47 +190,99 @@ created in the database directly via the `graphile_worker.add_job()` or
...
+}
+
+export const send_email: Task<"send_email"> = async function (payload) {
+- const { to, subject, body } = payload;
++ const { to, subject, body, from } = payload;
+  await ses.sendEmail({
+    Destination: {
+      ToAddress: [to],
+-      FromAddress: ["no-reply@example.com"],
++      FromAddress: [from ?? "no-reply@example.com"],
```
benjie, line 272:
```suggestion
      ToAddresses: [to],
-      FromAddresses: ["no-reply@example.com"],
+      FromAddresses: [from ?? "no-reply@example.com"],
```

website/docs/typescript.md
```diff
@@ -8,12 +8,146 @@ typed as `unknown` since they may have been populated by out of date code, or
...
+    return;
+  }
+  throw new Error("Invalid payload, expected a MyPayload");
+}
+
+const task: Task = async (payload) => {
+  assertMyPayload(payload);
+  console.log(payload.username);
+};
+```
```
benjie, line 39: I feel like this example may not be needed any more, or could be simplified?

benjie (changes requested):
Great start; I think this order definitely makes more sense.

The SES API uses plural keys so I've updated these in a few places. I think we should edit `assertMyPayload` to better match the `assertPayload` seen lower down (i.e. `if(blah)throw`) and also the error messages that we throw with should be more helpful - each one should be unique (rather than `"invalid"`).

website/docs/typescript.md
jemgillam: Yeah this is the original example, I feel like it should stay as a simple example at the top with the explanation. How can it be stripped down?

website/docs/typescript.md
```diff
@@ -56,47 +183,99 @@ created in the database directly via the `graphile_worker.add_job()` or
...
+      },
+    },
+  });
 };
 ```
 
-If this is too manual, you might prefer to use a library such as `runtypes` or
-the many others of a similar ilk. If you're not concerned with the type safety
-of the payload, you can work around it with a couple casts:
+If now we introduce the new functionality to set the from address, the changes
```
jemgillam, line 237:
```suggestion
If now we introduce the new functionality to set the `from` address, the changes
```

website/docs/typescript.md
```diff
@@ -56,47 +183,99 @@ created in the database directly via the `graphile_worker.add_job()` or
...
+we make have to take into account that older jobs may not have the from address
```
jemgillam, line 238:
```suggestion
we make have to take into account that older jobs may not have the `from` address
```

benjie (approved):
Excellent!

PR #413, opened 2023-12-13
Fix documentation on maxAttempts to match definition

okko:
## Description

The definition says maxAttempts minimum is 1 in which case the task will only be attempted once and won't be retried.

## Performance impact

None

## Security impact

None

## Checklist

- [X] My code matches the project's code style and `yarn lint:fix` passes.

### DIFF
--- a/website/docs/library/add-job.md
+++ b/website/docs/library/add-job.md
@@ -22,7 +22,7 @@ The `addJob` arguments are as follows:
 - `options`: an optional object specifying:
   - `queueName`: the queue to run this task under
   - `runAt`: a `Date` to schedule this task to run in the future
-  - `maxAttempts`: how many retries should this task get? (Default: 25)
+  - `maxAttempts`: the maximum number of attempts we'll give the job (Default: 25)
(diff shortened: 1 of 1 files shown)


benjie (dismissed):
Huh, indeed poor wording on my point - thanks! I think we could have changed `retries` to `tries` for the same effect, but this is fine too.

website/docs/library/add-job.md
```diff
@@ -22,7 +22,7 @@ The `addJob` arguments are as follows:
 - `options`: an optional object specifying:
   - `queueName`: the queue to run this task under
   - `runAt`: a `Date` to schedule this task to run in the future
-  - `maxAttempts`: how many retries should this task get? (Default: 25)
+  - `maxAttempts`: the maximum number of attempts we'll give the job (Default: 25)
```
benjie, line 25: Prettier...
```suggestion
  - `maxAttempts`: the maximum number of attempts we'll give the job
    (Default: 25)
```

benjie (approved):

PR #439, opened 2024-02-28
feat: add "cannot connect now" to list of retryable errors

psteinroe:
## Description

we experienced a downtime this morning after which the db was restarted. unfortunately, the queue was not restarted because "cannot connect now" is not retryable. this pr adds the respective error code to the list of retryable errors with a backoff time of 3s. retryable errors is now a list of object that holds the code and the backoff time

Edit: we have had another crash since this patch was applied and the queue recovered just fine. the crash was caused by a segfault.

## Performance impact

unknown, probably none.

## Security impact

unknown, probably none.

## Checklist


benjie (approved):
Looks good to me!

benjie:
Released in v0.16.3

PR #494, opened 2024-10-04
Docs: Fix add_job param order

dustin-H:
## Description

Hi there,
I think I found a mistake in the docs regarding `add_job` via SQL:
https://worker.graphile.org/docs/sql-add-job

In the top the order of parameters is described as `run_at` is just before `max_attempts`.

However, a little below in the "TIP" example it is used in reverse.
I think this is wrong so I fixed it.

## Performance impact
None

## Security impact
None

## Checklist

### DIFF
--- a/website/docs/sql-add-job.md
+++ b/website/docs/sql-add-job.md
@@ -83,8 +83,8 @@ SELECT graphile_worker.add_job(
   $1,
   payload := $2,
   queue_name := $3,
-  max_attempts := $4,
-  run_at := NOW() + ($5 * INTERVAL '1 second')
+  run_at := NOW() + ($4 * INTERVAL '1 second'),
+  max_attempts := $5
 );
 ```
 


benjie:
Hi Dustin! Thanks for the contribution, but with named parameters the order is unimportant - this is one of the main reasons to use named parameters.

dustin-H:
Hi Benjie,
ahh right. Sorry for coming up with this.

I came across this, as I copied this query to my project and removed the parameter names without thinking about it. This lead to a bug obviously.

So, sorting it could prevent that for other.

But the decision is on you. Just close this PR if you don't think so.

Thank you!

benjie:
The example above this one is:

```sql
```

That would definitely not be valid if you removed `run_at :=`! I think we'll leave it as is, but thanks for bringing this to my attention - if someone reports a similar issue I'll be sure to ask them if they've removed any parameter names!

PR #497, opened 2024-10-16
Expand on considerations re maxPoolSize

eyn:
## Description
Add an explanation around maxPoolSize based on discord conversation.

## Performance impact
N/A

## Security impact
N/A

## Checklist
N/A - documentation only

### DIFF
--- a/website/docs/config.md
+++ b/website/docs/config.md
@@ -144,9 +144,9 @@ A Logger instance.
 Type: `number | undefined`
 
 Maximum number of concurrent connections to Postgres. This number can be lower
-than concurrentJobs. However a low pool size may cause you issues. If all your 
-pool clients are busy then no jobs can be started and no jobs can be 
-released, so it's critical that you don't run out.
+than concurrentJobs. However a low pool size may cause you issues. If all your
+pool clients are busy then no jobs can be started and no jobs can be released,
+so it's critical that you don't run out.
 
 ### worker.maxResetLockedInterval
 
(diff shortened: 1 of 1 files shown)


website/docs/config.md
```diff
@@ -143,7 +143,10 @@ A Logger instance.
 
 Type: `number | undefined`
 
-Maximum number of concurrent connections to Postgres
+Maximum number of concurrent connections to Postgres. This number can be lower
+than concurrentJobs. However a low pool size may cause you issues. If all your
+pool clients are busy then no jobs can be started and no jobs can be released,
+so it's critical that you don't run out.
```
benjie, line 149:
```suggestion
so it's critical that you don't run out. It must be at least `2`.
```

website/docs/config.md
```diff
@@ -143,7 +143,10 @@ A Logger instance.
 
 Type: `number | undefined`
 
-Maximum number of concurrent connections to Postgres
+Maximum number of concurrent connections to Postgres. This number can be lower
+than concurrentJobs. However a low pool size may cause you issues. If all your
+pool clients are busy then no jobs can be started and no jobs can be released,
+so it's critical that you don't run out. It must be at least `2`.
```
benjie, line 149:
```suggestion
Maximum number of concurrent connections to Postgres; must be at least `2`. This
number can be lower than `concurrentJobs`, however a low pool size may cause
issues - if all your pool clients are busy then no jobs can be started or
released. If in doubt, we recommend setting it to `10` or `concurrentJobs + 2`,
whichever is larger. (Note: if your task executors use this pool, then an even
larger value may be needed for optimum performance, depending on the shape of
your logic.)
```

benjie (approved):
Awesome; thanks!

PR #499, opened 2024-10-31
Update doc ('pattern' -> 'match')

jrr:
The 0.14 release renamed "pattern" to "match", and the doc didn't get the memo.

It looks like the new "match" field can also take a `CronMatcher`, but I'm not sure how that works.

https://github.com/graphile/worker/blob/414b11cb2a7eff0459462fd5d788d821b33f406b/src/interfaces.ts#L370

### DIFF
--- a/website/docs/cron.md
+++ b/website/docs/cron.md
@@ -232,8 +232,8 @@ has the following properties:
   (same as the first argument to `add_job`)
 - `match` (required): a cron pattern (e.g. `* * * * *`) describing when to run,
   or a callback function passed a TimestampDigest and returning a boolean
-  whether the cron should fire for this timestamp (true) or not (false)
-  this task
+  whether the cron should fire for this timestamp (true) or not (false) this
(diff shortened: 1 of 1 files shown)


website/docs/cron.md
```diff
@@ -230,7 +230,7 @@ has the following properties:
 
 - `task` (required): the string identifier of the task that should be executed
   (same as the first argument to `add_job`)
-- `pattern` (required): a cron pattern (e.g. `* * * * *`) describing when to run
+- `match` (required): a cron pattern (e.g. `* * * * *`) describing when to run
```
benjie, line 233:
```suggestion
- `match` (required): a cron pattern (e.g. `* * * * *`) describing when to run,
  or a callback function passed a TimestampDigest and returning a boolean
  whether the cron should fire for this timestamp (true) or not (false)
```

benjie (approved):
Thanks!

Issue #495, opened 2024-10-06
When specifying `CronItemOptions`, `backfillPeriod` is not optional in TypeScript


danieldiekmeier:
### Summary

I'm defining my cronList as an array of `CronItem`s, so that I can later pass them through `parseCronItems`. This has worked great! (Thank you very much for the library!)

Today, I wanted to add `maxAttempts: 1` to one of my `CronItem`s, and created an `options` object for that. TypeScript started complaining that the options must have a `backfillPeriod` key.

### Steps to reproduce

Use this code:

```ts
import { type CronItem } from 'graphile-worker'

const cronList: CronItem[] = [
  {
    task: 'example',
    match: '* * * * *,
    identifier: 'example',
    options: {
      maxAttempts: 1,
      // TypeScript: Property 'backfillPeriod' is missing in type '{ maxAttempts: number; }' 
      //             but required in type 'CronItemOptions'. ts(2741)
    },
  },
}
```

### Expected results

I'd expect the key to be optional, like the other options in `CronItemOptions`.

### Actual results

At runtime, this seems to work correctly and I don't see any issues. The problem is only the type error.

### Additional context

- Node.js 20.17.0
- TypeScript v5.6.2

### Possible Solution

I think the problem is that Graphile Worker uses the same interface internally and externally. If I understand correctly, the `backfillPeriod` is defaulted to 0 here:

https://github.com/graphile/worker/blob/436e29968d28c48d5f46a012b9c83b0747e2eff1/src/crontab.ts#L151-L153

So it makes sense that _internally_, `backfillPeriod` is always defined. But since this runtime default exists, I think there should be a slightly different interface to the outside.

PR #516, opened 2024-12-03
make backfillPeriod optional in exported CronItem type

jcgsville:
## Description

Addresses https://github.com/graphile/worker/issues/495

This allows users to not specify a backfill period, and worker will default it to zero. To support this, `parseCronItem()` also now defaults it to zero in addition to `parseCrontabOptions()`

This PR changes the external interface, but it should be backwards-compatible with existing usage.

## Performance impact

I don't expect there would be any performance impact. There is almost no runtime code changed.

## Security impact

Similarly, there should not be any impact to security considerations.

## Checklist

- [x] ~I have detailed the new feature in the relevant documentation.~ I do not plan to add this to documentation.

### DIFF
--- a/__tests__/crontab.test.ts
+++ b/__tests__/crontab.test.ts
@@ -1,4 +1,4 @@
-import { $$isParsed, CronItemOptions, ParsedCronMatch } from "../src";
+import { CronItemOptions, ParsedCronMatch } from "../src";
 import { parseCronItem, parseCrontab } from "../src/crontab";
 
 // 0...59
@@ -250,7 +250,7 @@ describe("parses JS cron items correctly", () => {
   });
 
   test("validates match strings", () => {
-    const matchString = "foobar";
+    const matchString = "foo";
     expect(() =>
       parseCronItem({
         task: "task",
--- a/__tests__/crontab.test.ts
+++ b/__tests__/crontab.test.ts
@@ -257,8 +257,6 @@ describe("parses JS cron items correctly", () => {
         match: matchString,
         options: {},
       }),
-    ).toThrow(
-      `"Invalid cron pattern '${matchString}' in parseCronItem call"`,
-    );
+    ).toThrow(`Invalid cron pattern '${matchString}' in parseCronItem call`);
   });
 });
--- a/__tests__/crontab.test.ts
+++ b/__tests__/crontab.test.ts
@@ -257,7 +257,7 @@ describe("parses JS cron items correctly", () => {
         match: matchString,
         options: {},
       }),
-    ).toThrowErrorMatchingInlineSnapshot(
+    ).toThrow(
       `"Invalid cron pattern '${matchString}' in parseCronItem call"`,
     );
   });
--- a/__tests__/crontab.test.ts
+++ b/__tests__/crontab.test.ts
@@ -1,5 +1,5 @@
-import { CronItemOptions, ParsedCronMatch } from "../src";
-import { parseCrontab } from "../src/crontab";
+import { $$isParsed, CronItemOptions, ParsedCronMatch } from "../src";
+import { parseCronItem, parseCrontab } from "../src/crontab";
 
 // 0...59
 const ALL_MINUTES = Array.from(Array(60).keys());
@@ -234,6 +234,31 @@ describe("gives error on syntax error", () => {
 `),
     ).toThrowErrorMatchingInlineSnapshot(
       `"Failed to parse JSON5 payload on line 1 of crontab: JSON5: invalid character '=' at 1:13"`,
+    );
+  });
+});
+
+describe("parses JS cron items correctly", () => {
+  test("defaults backfillPeriod to 0", () => {
+    expect(
+      parseCronItem({
+        task: "task",
+        match: "* * * * *",
+        options: {},
+      }).options.backfillPeriod,
+    ).toEqual(0);
+  });
+
+  test("validates match strings", () => {
+    const matchString = "foobar";
+    expect(() =>
+      parseCronItem({
+        task: "task",
+        match: matchString,
+        options: {},
+      }),
+    ).toThrowErrorMatchingInlineSnapshot(
+      `"Invalid cron pattern '${matchString}' in parseCronItem call"`,
     );
   });
 });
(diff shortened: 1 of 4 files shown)


jcgsville:
I went to go update the documentation, and I realized I don't actually know the use case for importing CronItem specifically instead of ParsedCrontItem.

It looks like parseConItem and parseCronItems are both exported in `src/index.ts`, so if you wanted to use those functions, you might benefit from this change, but I'm also not sure why you would use those functions.

@danieldiekmeier, can you elaborate what your use case was so that I might appropriately update the documentation?

src/interfaces.ts
```diff
@@ -326,10 +326,14 @@ export interface WatchedCronItems {
   release: () => void;
 }
 
+export type CronItemOptions = {
+  [key in keyof ParsedCronItemOptions]?: ParsedCronItemOptions[key];
+};
```
benjie, line 331: Does this maintain the documentation for these fields in TypeScript hover/etc?

src/interfaces.ts
```diff
@@ -326,10 +326,14 @@ export interface WatchedCronItems {
...
+
 /**
  * a.k.a. `opts`, this allows you to change the behaviour when scheduling a cron task.
  */
-export interface CronItemOptions {
+export interface ParsedCronItemOptions {
   /** How far back (in milliseconds) should we backfill jobs when worker starts? (Only backfills since when the identifier was first used.) */
   backfillPeriod: number;
```
benjie, line 338: Instead, make the field optional:

```suggestion
export interface CronItemOptions {
  /** How far back (in milliseconds) should we backfill jobs when worker starts? (Only backfills since when the identifier was first used.) */
  backfillPeriod?: number;
```

and then later make it explicit just this one field is being changed internally:

```ts
export interface ParsedCronItemOptions extends CronItemOptions {
  backfillPeriod: number;
}
```

danieldiekmeier:
Yes, that is exactly my use case. I configure Graphile Worker programmatically and this is basically my setup:

```ts
import {
  parseCronItems,
  run,
  type CronItem
} from 'graphile-worker'

const cronList: CronItem[] = [
  {
    task: 'enqueuePostalDispatches',
    match: HOURLY,
    identifier: 'hourlyPostalDispatch',
    options: {
      backfillPeriod: 0, // Currently required: https://github.com/graphile/worker/issues/495
      maxAttempts: 1,
    },
  },
  {
    task: 'deleteOldTemporaryPdfs',
    match: '5 4 * * *', // daily at 04:05
    identifier: 'dailyPdfDeleter',
    options: {
      backfillPeriod: 0, // Currently required: https://github.com/graphile/worker/issues/495
      maxAttempts: 1,
    },
  },
]

await run({
  pgPool: pool,
  concurrency: 2,
  taskList,
  parsedCronItems: parseCronItems(cronList),
})
```

My current workaround is to explicitly set `backfillPeriod: 0`, even though at runtime it is set to 0 either way. I only do it to appease TypeScript.

src/crontab.ts
```diff
@@ -308,11 +309,24 @@ export const parseCronItem = (
...
+  if (!options) {
+    parsedOptions = { backfillPeriod: 0 };
+  } else if (options.backfillPeriod === undefined) {
+    parsedOptions = { ...options, backfillPeriod: 0 };
+  } else {
+    parsedOptions = {
+      ...options,
+      backfillPeriod: options.backfillPeriod,
+    };
+  }
```
benjie, line 323: I like to avoid the use of `let` where possible because it opens the door to reassignment; please use a ternary instead... Though that's not necessary here:

```suggestion
  const parsedOptions: ParsedCronItemOptions = {
    ...options,
    backfillPeriod: options?.backfillPeriod ?? 0,
  };
```

jcgsville:
@benjie, This impacts the input type to `parseItems()`, which I think is compatible with that section of the docs you linked, right? People shouldn't be using `ParsedCronItem`, but using `CronItem` is ok.

Thanks for chiming in, @danieldiekmeier. Based on what I'm seeing, I don't think this change warrants an addition to the documentation. Happy to add somewhere though if you think otherwise, Benjie

src/crontab.ts
jcgsville: That's embarrassingly much better than my code 😅 Thanks for the suggestion. Applied

src/interfaces.ts
jcgsville: Screenshot 2024-12-04 at 5 21 17 PM

I believe so

src/interfaces.ts
jcgsville: This seems to be more inline with the existing style of using interfaces over types. I'll adjust it to this.

jcgsville:
Also, just fyi @danieldiekmeier, if I understand the code correctly, `parseCronItems()` doesn't actually set it to zero by default right now. Only `parseCrontabOptions()` does which parses the option strings, not the JS options. So until you get this change in your code, best to continue passing that zero in. I think it's more than just to appease TS.

__tests__/crontab.test.ts
```diff
@@ -237,3 +237,28 @@ describe("gives error on syntax error", () => {
...
+
+  test("validates match strings", () => {
+    const matchString = "foo";
+    expect(() =>
+      parseCronItem({
+        task: "task",
+        match: matchString,
+        options: {},
+      }),
+    ).toThrowErrorMatchingInlineSnapshot(
```
benjie, line 260: An inline snapshot can but updated by `jest -u`; but yours includes a variable? 
```suggestion
    ).toThrow(
```
https://jestjs.io/docs/expect#tothrowerror

benjie (dismissed):
Looks good from a quick scan; will review more thoroughly later.

danieldiekmeier:
@jcgsville I also don't think it needs changes to the docs, since the docs already claim that it works that way: https://worker.graphile.org/docs/cron#parsedcronitems, where it says "`options`: optional options influencing backfilling, etc".

I didn't mean that `parseCronItems` would set the setting to 0 at runtime. It happens somewhere deeper in the runner. But I think we don't need to discuss the exact reason why I'll leave this line in the code :D

benjie (approved):
Looks good to me! :raised_hands:

__tests__/crontab.test.ts
jcgsville: Shoot. Careless copilot autocomplete. Thanks for addressing

Issue #453, opened 2024-04-09
Documentation: detail how to pass a preset to `run()`


benjie:
```ts
import type {} from "graphile-config";
import { run } from "graphile-worker";
import { WorkerProPreset } from "@graphile-pro/worker";

const preset: GraphileConfig.Preset = {
  extends: [WorkerProPreset],
  worker: {
    /* Your worker config options here */
    connectionString: process.env.DATABASE_URL,
    maxPoolSize: 10,
    // ...
  },
};

const promise = run({ preset });
```

This needs detailing both on the library page and the worker pro page.

jcgsville:
Just stumbled over this, and it took me a couple minutes to figure this out. Are you open to me adding it to the docs, or do you want to save it for another contributor for their first issue?

benjie:
Please have at it! Presets are very new in Worker, there's a lot more to be done with their integration (and the deprecation of the non-preset version). I'm trying to standardize the Graphile suite around Graphile Config so that all of our software uses similar patterns for plugins, presets, middleware, hooks, etc - but Graphile Config itself is currently incredibly poorly documented because it's part of the PostGraphile V5 system and is itself still being iterated.

jcgsville:
Coolio. Happy to also take a stab at improving the documentation for Graphile Config, though it sounds like that may want to wait until it's more stable?

(I'm viewing doing a pass on documentation as a good way to understand more deeply how these things work with the goal of enabling both ongoing contribution and better understanding for my future usage of graphile stuff)

benjie:
It’s got quite wide usage already, so documenting the commonly used parts at least would be wise. I think Graphile config needs two docs: one for library authors/plugin authors, and one for regular consumers.

PR #520, opened 2024-12-05
update config docs

jcgsville:
## Description

A bunch of small updates to the Configuration docs. Wording tweaks for clarity, additions to fill in gaps and make things more consistent across the CLI, TS, and markdown docs.

Partially resolves https://github.com/graphile/worker/issues/453. Should probably update the /library/run docs as well. I'll do so when I get to that page.

## Performance impact

None

## Security impact

None

## Checklist

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -56,7 +56,7 @@ const argv = yargs
   .number("poll-interval")
   .option("no-prepared-statements", {
     description:
-      "set this flag if you want to disable prepared statements, e.g. for compatibility with pgBouncer",
+      "set this flag if you want to disable prepared statements, e.g. for compatibility with pgBouncer < 1.21.0",
   })
   .boolean("no-prepared-statements")
   .option("config", {
--- a/src/index.ts
+++ b/src/index.ts
@@ -64,7 +64,7 @@ declare global {
        */
       connectionString?: string;
       /**
-       * Maximum number of concurrent connections to Postgres
+       * Maximum number of concurrent connections to Postgres. Must be at least `2`.
        *
        * @defaultValue `10`
        */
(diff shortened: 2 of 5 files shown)


src/cli.ts
```diff
@@ -56,7 +56,7 @@ const argv = yargs
   .number("poll-interval")
   .option("no-prepared-statements", {
     description:
-      "set this flag if you want to disable prepared statements, e.g. for compatibility with pgBouncer",
+      "set this flag if you want to disable prepared statements, e.g. for compatibility with pgBouncer < 1.21.0",
```
benjie, line 59:
```suggestion
      "set this flag if you want to disable prepared statements, e.g. for compatibility with some external PostgreSQL pools",
```

src/index.ts
```diff
@@ -73,86 +73,100 @@ declare global {
        *
        * @defaultValue `2000` */
       pollInterval?: number;
-      /** @defaultValue `true` */
+      /**
+       * Whether Graphile Worker should use prepared statements. Set to `false` for
+       * compatibility with pgBouncer < 1.21.0.
```
benjie, line 78:
```suggestion
       * Whether Graphile Worker should use prepared statements. Set `false` if you
       * use software (e.g. some Postgres pools) that don't support them.
```

src/index.ts
```diff
@@ -73,86 +73,100 @@ declare global {
...
+       *
+       * @defaultValue `true`
+       */
       preparedStatements?: boolean;
       /**
-       * The database schema in which Graphile Worker is (to be) located.
+       * The database schema in which Graphile Worker's tables, functions, views, etc are
+       * located. Database migrations will create or edit things in this schema
+       * if necessary.
```
benjie, line 86:
```suggestion
       * located. Graphile Worker will create or edit things in this schema as
       * necessary.
```

src/index.ts
```diff
@@ -73,86 +73,100 @@ declare global {
...
       /**
-       * Override path to crontab file.
+       * The path to a file in which Graphile Worker should look for crontab schedules.
        *
        * @defaultValue `process.cwd() + "/crontab"`
        */
       crontabFile?: string;
       /**
-       * Number of jobs to run concurrently.
+       * Number of jobs to run concurrently on a single worker.
```
benjie, line 104:
```suggestion
       * Number of jobs to run concurrently on a single Graphile Worker instance.
```

src/index.ts
```diff
@@ -73,86 +73,100 @@ declare global {
...
        * **Experimental**
        *
-       * How often should we scan for jobs that have been locked too long and
-       * release them? This is the minimum interval, we'll choose a time between
-       * this and `maxResetLockedInterval`.
+       * How often should Graphile Worker scan for and release jobs that have been
+       * locked too long? This is the minimum interval in milliseconds. Graphile
+       * Worker will choose a time between this and `maxResetLockedInterval`.
+       *
+       * @defaultValue `480_000`
```
benjie, line 139: Keeping this correct is a maintenance burden, do we really need it? Can we provide it another way such that there's only one source of truth?

src/index.ts
```diff
@@ -73,86 +73,100 @@ declare global {
...
        */
       minResetLockedInterval?: number;
       /**
        * **Experimental**
        *
-       * The upper bound of how long we'll wait between scans for jobs that have
-       * been locked too long. See `minResetLockedInterval`.
+       * In milliseconds, the upper bound of how long Graphile Worker will wait
```
benjie, line 145:
```suggestion
       * The upper bound of how long (in milliseconds) Graphile Worker will wait
```

src/index.ts
```diff
@@ -73,86 +73,100 @@ declare global {
...
+       * between scans for jobs that have been locked too long (see
+       * `minResetLockedInterval`).
+       *
+       * @defaultValue `600_000`
        */
       maxResetLockedInterval?: number;
-
```
benjie, line 141: Please restore all of these blank newlines (and add any missing ones) - I use `{` and `}` in vim to navigate the code via blank lines; it also makes it easier to read the code - knowing that the comment applies to the entry below it rather than above it.

src/index.ts
```diff
@@ -73,86 +73,100 @@ declare global {
...
       /**
        * **Experimental**
        *
-       * When getting a queue name in a job, we batch calls for efficiency. By
-       * default we do this over a 50ms window; increase this for greater efficiency,
-       * reduce this to reduce the latency for getting an individual queue name.
+       * The window size in milliseconds in which Graphile Worker batches calls for
+       * getting a queue name in a job. This batching is done for efficiency. Increase
+       * this window for greater efficiency. Reduce this window to reduce the
+       * latency for getting an individual queue name.
```
benjie, line 158:
```suggestion
       * The size, in milliseconds, of the time window over which Graphile Worker
       * will batch requests to retrieve the queue name of a job. Increase the size
       * of this window for greater efficiency, or reduce it to improve latency.
```

src/index.ts
```diff
@@ -73,86 +73,100 @@ declare global {
...
        */
       getQueueNameBatchDelay?: number;
-
       /**
        * A Logger instance.
        */
       logger?: Logger;
-
+      /**
+       * A Node.js `EventEmitter` that exposes certain events within the runner.
```
benjie, line 168:
```suggestion
       * Provide your own Node.js `EventEmitter` in order to be able to receive events
       * that occur during Graphile Worker's startup. (Without this, Worker will
       * provision its own `EventEmitter`, but you can't retrieve it until the promise
       * returned by the API you have called has resolved.)
```

src/interfaces.ts
```diff
@@ -669,7 +669,7 @@ export interface SharedOptions {
 
   /**
    * Set true if you want to prevent the use of prepared statements; for
-   * example if you wish to use Graphile Worker with pgBouncer or similar.
+   * example if you wish to use Graphile Worker with pgBouncer < 1.21.0 or similar.
```
benjie, line 672:
```suggestion
   * example if you wish to use Graphile Worker with a connection pool
   * that does not support prepared statements.
```

website/docs/cli/run.md
```diff
@@ -57,7 +57,7 @@ Options:
                                 future/retries)         [number] [default: 2000]
       --no-prepared-statements  set this flag if you want to disable prepared
                                 statements, e.g. for compatibility with
-                                pgBouncer             [boolean] [default: false]
+                                pgBouncer < 1.21.0    [boolean] [default: false]
```
benjie, line 60: This should auto-update

benjie (changes requested):
Definitely good to clarify a lot of this! I'm filing this review so that I can switch to direct editorial, doing this all in GitHub comments is becoming unwieldy.

benjie:
Please rebase (or merge `main` if it's easier - I don't mind which) now https://github.com/graphile/worker/pull/521 has been addressed. Turns out some of the docs updates were merged into the wrong place.

benjie (approved):
Massive improvement - thanks! :raised_hands:

src/index.ts
jcgsville: Thanks for updating this one. Still not exactly sure what this one does. Got it on my todo list to dive in and figure out :)

src/index.ts
benjie: Currently Graphile Config doesn't even know about the subprojects that use it - there's no "registry" or similar - but I don't think we'll keep it isolated like that long term... I think we'll need to know how to merge certain options (should we: overwrite, concatenate, throw, etc?). So maybe if we do that codegen might make sense.

For now, I'm okay with people looking up the defaults in the source and/or docs. Or putting it in the prose to indicate that it might have changed but it was this at one point.

Or just expecting the community to send PRs when they notice the comments are out of sync :wink:

jcgsville: By "one place", I meant that each setting's default should be defined in one place, not necessarily one registry to rule them all.

But, if codegen is on the table in the future, then maybe there is a world where graphile config does have a preset registry. That could be interesting.

But, I don't know if I love having the preset *definition* own the default 🤔 It seems like the one of the points of presets is you can somewhat easily share different configurations / different defaults, and customize from there.

Side note, it would be pretty nice to encode more validation than just the TS type in the preset definition. I'm thinking about how one maxPoolSize has to be at least 2, for example. Maybe an opportunity to incorporate a zod schema for a preset definition.

jcgsville:
I noticed you took out this bit

```
Adding the import statement tells TypeScript about the `GraphileConfig` global
namespace and the properties that
[Graphile Worker adds](https://github.com/graphile/worker/blob/c70e9db03f6ad292dcdb833714741363bd78937d/src/index.ts#L158)
to the `Preset` interface. No code from the `graphile-worker` or
`graphile-config` libraries will be included in the output JavaScript for
`graphile.config.ts` above. See the TypeScript docs for more info about
[declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html)
and
[type-only imports](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export).
```

Do you think that was a distraction / unnecessary? I thought it might help people to understand how the graphile config stuff works. Maybe better included elsewhere?

src/index.ts
benjie: All good ideas, but I’m going to delay thinking about them until v6 ;)

benjie:
It was good; but when I ran the `graphile config print` command nothing much was output and I realised it was because we didn’t explicitly add the worker preset to the config (because it’s added for us). By adding the preset explicitly the print command worked and there was no need for the stub import type, which meant there was nothing for the note to relate to.

benjie:
Definitely consider adding something like it to the graphile config docs though

jcgsville:
Ah, makes sense. Sounds good 👍

jcgsville:
added the stuff about the graphile config global namespace in my latest graphile config docs pr:

https://github.com/graphile/crystal/pull/2282

PR #522, opened 2024-12-06
tweaks to tasks documentation for clarity and consistency

jcgsville:
Also including a couple small followups to the config documentation tweaks

### DIFF
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -181,8 +181,8 @@ export interface JobHelpers extends Helpers {
   ): Promise<QueryResult<R>>;
 
   /**
-   * A Node `AbortSignal` that will be triggered when the job should exit. It is
-   * used, for example, for a graceful shutdown request.
+   * An `AbortSignal` that will be triggered when the job should exit. It is used,
+   * for example, for a graceful shutdown request.
    *
    * @experimental
    */
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -134,7 +134,7 @@ export type AddJobsFunction = <TSpecs extends readonly AddJobsJobSpec[]>(
 
 export interface Helpers {
   /**
-   * A Logger instance.
+   * A `Logger` instance.
    */
   logger: Logger;
 
@@ -145,30 +145,30 @@ export interface Helpers {
   withPgClient: WithPgClient;
 
   /**
-   * Adds a job into our queue.
+   * Adds a job into the Graphile Worker queue.
    */
   addJob: AddJobFunction;
 
   /**
-   * Adds multiple jobs into our queue.
+   * Adds multiple jobs into the Graphile Worker queue.
    */
   addJobs: AddJobsFunction;
 }
 
 export interface JobHelpers extends Helpers {
   /**
-   * A Logger instance, scoped to this job.
+   * A `Logger` instance, scoped to this job.
    */
   logger: Logger;
 
   /**
-   * The currently executing job.
+   * The whole, currently executing job.
    */
   job: Job;
 
   /**
-   * Get the queue name of the give queue ID (or the currently executing job if
-   * no queue id is specified).
+   * Get the queue name of the given queue ID (or of the currently executing job
+   * if no queue ID is specified).
    */
   getQueueName(queueId?: number | null): PromiseOrDirect<string | null>;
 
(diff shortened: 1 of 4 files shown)


src/interfaces.ts
```diff
@@ -181,7 +181,8 @@ export interface JobHelpers extends Helpers {
   ): Promise<QueryResult<R>>;
 
   /**
-   * An AbortSignal that will be triggered when the job should exit.
+   * A Node `AbortSignal` that will be triggered when the job should exit. It is
```
benjie, line 184: You say a Node one, but AbortSignal is a DOM standard that Node implements I think? Not sure this needs specifying since there’s no ambiguity?

website/docs/tasks.md
```diff
@@ -82,44 +66,39 @@ module.exports = async (payload, helpers) => {
 };
 ```
 
-## The `tasks/` folder
+## The task directory
```
benjie, line 69:
```suggestion
## The tasks directory
```

benjie: Hmmm I see why you made this change. I wonder why I named the variable that.

src/interfaces.ts
jcgsville: This is a good point. I think I first added the word Node before I added the extra bit of clarity to the dm. I think it's fine without the word "Node"

website/docs/tasks.md
```diff
@@ -130,16 +109,15 @@ TypeScript files to JS and then have Graphile Worker load the JS files.
...
+1. Install `ts-node`.
+2. Add `".ts"` to the `worker.fileExtensions` list in your preset.
+3. Run Graphile Worker with the environment variable
    `NODE_OPTIONS="--loader ts-node/esm"` set.
 
 ```ts title="Example graphile.config.ts"
-import type { GraphileConfig } from "graphile-config";
 import type {} from "graphile-worker";
 
 const preset: GraphileConfig.Preset = {
```
benjie, line 123: We should `extends` the WorkerPreset here like we do on the config page

website/docs/tasks.md
```diff
@@ -167,16 +145,16 @@ This feature is currently experimental.
...
-setting the relevant environmental variables and passing in the payload
+finds an executable file inside of the `taskDirectory` it will create a task
+executor for it. When a task of this kind is found, Graphile Worker will execute
+the file with the relevant environment variables and pass in the payload
 according to the encoding. If the executable exits with code `0` then Graphile
-Worker will see this as success, all other exit codes are seen as failure.
+Worker will see this as success. All other exit codes are seen as failure.
 
-### Environmental variables
+### Environment variables
```
benjie, line 154: Wow; I’ve been wrong about this my entire career 🤯

benjie (approved):
Looks good to me! Will do a final review/merge next week. Thanks! 🙏

website/docs/tasks.md
jcgsville: I had to google this to see if this is a UK thing or something 🙂

website/docs/tasks.md
```diff
@@ -236,9 +241,10 @@ This is a convenience wrapper for
 
 ### `helpers.withPgClient()`
 
-`withPgClient` gets a `pgClient` from the pool, calls
-`await callback(pgClient)`, and finally releases the client and returns the
-result of `callback`. This workflow can make testing your tasks easier.
+`withPgClient` gets a `pgClient` from the pool that Graphile Worker uses. It
+calls `await callback(pgClient)`, and finally releases the client and returns
+the result of `callback`. This workflow can make testing your tasks easier by
+making it easier to mock `pgClient`.
```
benjie, line 248: One thing to know about me is I almost never use mocks. How it makes testing easier for me is I always return the same client, which can be in a transaction, so I can roll back the changes later; however I’d be careful about how that’s expressed here because I wouldn’t want a user thinking it will always return the same client normally.

jcgsville: Gotcha. I think I'll revert this addition for now. When I first read that it would make testing easier, my next question was "why?". Was trying to avoid that, but I think this level of nuance is probably better communicated in a future section on testing, or something like that

benjie (approved):
Looks great; thanks!

benjie:
This is now live :tada:

PR #553, opened 2025-07-20
WIP: Use new PostgreSQL adapter pattern

### LINKED ISSUE #441: postgres.js pool support
### Feature description

I think Worker should ideally be able to accept a pool from either node-postgres or postgres.js

### Motivating example

I am currently working on a project that connects to the db using postgres.js, my app does not use node-postgres anywhere, and ideally I shouldn't be required to add another adapter.

### Breaking changes

I don't believe this should constitute a breaking change

### Supporting development

I [tick all that apply]:

- [x] am interested in building this feature myself
- [x] am interested in collaborating on building this feature
- [x] am willing to help testing this feature before it's released
- [x] am willing to write a test-driven test suite for this feature (before it exists)
- [x] am a [Graphile sponsor](https://www.graphile.org/sponsor/) ❤️
- [ ] have an active [support or consultancy contract](https://www.graphile.org/support/) with Graphile

leo91000:
## Summary

  This PR updates graphile-worker to use the new PostgreSQL adapter pattern from @graphile/pg-core, addressing #441. The adapter repository code is available here : https://github.com/leo91000/graphile_pg/tree/feat/stream-first-adapters

  **⚠️ This is a work in progress:**
  - Some tests are still failing (unrelated to adapter changes - Jest module loading issues)
  - The adapter API might still change

  ## Changes

  ### 1. PostgreSQL Adapter Pattern
  - Replaced direct `pg` module usage with adapter interfaces (`PgPool`, `PgClient`)
  - Made the adapter optional - it's dynamically imported only when needed
  - Moved `pg`, `@types/pg`, and `@graphile/pg-adapter-node-postgres` to devDependencies

  ### 2. Connection String Handling

  **Before (v0.16.x):**
  ```js
  // Worker creates a pg.Pool internally
  await run({
    connectionString: 'postgres://...'
  });
```

  After (with this PR):
  ```js
  // Option 1: Let worker create pool (requires adapter packages installed)
  await run({
    connectionString: 'postgres://...'
  });
  // This will dynamically import @graphile/pg-adapter-node-postgres

  // Option 2: Create your own pool (recommended)
  import { createNodePostgresPool } from '@graphile/pg-adapter-node-postgres';
  const pgPool = createNodePostgresPool({ connectionString: 'postgres://...' });
  await run({ pgPool });
```

 ### 3. Breaking Changes

  Users who rely on connection strings must install the adapter packages:
  npm install @graphile/pg-adapter-node-postgres pg

  If these packages are not installed, they'll get a helpful error message guiding them to install the required packages.

 ### 4. Benefits

  - Flexibility: Users can choose their PostgreSQL client implementation
  - Better control: Users who provide their own pool maintain full control over its lifecycle
  - Backward compatibility: Connection strings still work with the adapter packages installed

  Test Status

  - ✅ 16/23 test suites passing
  - ❌ 7 test suites failing due to Jest dynamic import issues (not related to adapter changes)

  Related Issues

  Fixes #441


benjie:
This looks promising! However, it doesn't seem to quite align with the adaptors used in `@dataplan/pg`, and rewriting those will be significantly more work (especially because people are already using that adaptor interface to support other clients/ORMs/etc) - I'd rather stick with that design unless there's a very good reason to change it, because I really really really want to release PostGraphile V5 and I'd rather avoid any further delays.

I'd advise that rather than creating new packages that we then have to integrate with both projects, you instead create this as a PR to Graphile Crystal that extracts the logic from `@dataplan/pg` into another package, and ensures it still passes all of the tests. There will likely be a little complexity in this, some assumptions based around the @dataplan/pg use case that aren't true of the Graphile Worker use case, but I'm sure these can be addressed (and I'm happy to offer assistance there). Hopefully a lot of the work you've done in this can be ported fairly straightforwardly.

Ideally I'd like to make it so the same `graphile.config.ts` file could be used by both projects, and they can use the same adaptors, but likely with different configuration. (I don't think Worker would use the `preset.pgServices` that PostGraphile uses, instead maybe `preset.worker.dbPool` or similar.)

I would rather that these adaptors use the "subpath" approach that we use for the server adaptors, rather than the multiple modules approach:

https://github.com/graphile/crystal/blob/951dff78f50ff770679a0d10de317163a3088672/grafast/grafserv/package.json#L13-L48

The subpath approach means the user needs to put in less effort to use it, they can just do `yarn add graphile-worker postgres` and then start using the postgres adaptor, rather than also having to know the name of our adaptor.

leo91000:
Thank you for the feedback ! Will start working on it next week !

PR #556, opened 2025-07-29
chore: update graphile-config, tweak TS setup, Node20+ min

benjie:
Having `graphile.config.ts` have to import from `./dist/index.js` was annoying me :man_shrugging:


benjie:
But not enough to work through actually fixing it :disappointed:

PR #558, opened 2025-08-13
docs: improve queueName docs to avoid high cardinality

blimmer:
## Description

We recently ran into an issue where our serverless database had to dramatically scale up capacity. Upon reviewing those queries, we realized that the issue was caused by Graphile Worker query:

```sql
with j as ( 
    select jobs.job_queue_id, jobs.priority, jobs.run_at, jobs.id 
    from graphile_worker . _private_jobs
 as jobs 
    where jobs.is_available = ? and run_at <= now ( ) and task_id = any ( ? :: int [ ] ) and ( jobs.job_queue_id is ? or jobs.job_queue_id in ( 
        select id 
        from graphile_worker . _private_job_queues
 as job_queues 
        where job_queues.is_available = ? for 
        update skip locked 
    ) ) 
order by priority asc, run_at asc 
limit ? for 
update skip locked 
), q as ( 
    update graphile_worker . _private_job_queues
 as job_queues 
    set locked_by = ? :: text, locked_at = now ( ) 
    from j 
    where job_queues.id = j.job_queue_id 
) 
update graphile_worker . _private_jobs
 as jobs 
set attempts = jobs.attempts + ?, locked_by = ? :: text, locked_at = now ( ) 
from j 
where jobs.id = j.id returning *
```

Specifically the subquery against the `_private_job_queues` table.

```
select id from graphile_worker._private_job_queues
  where job_queues.is_available = ?
  for update skip locked
```

We were creating lots of job queues using high-cardinality values (e.g., database row IDs), which caused many thousands of dead queues to build up in this table.

```shell
db=> select count(*) from graphile_worker._private_job_queues;
 count
-------
 61106
(1 row)
```

Because the graphql worker polling job necessarily does a table scan on `_private_job_queues`, this caused a big performance issue for us, forcing the DB to scale up.

When I reviewed the documentation, I found a warning about not using randomly-generated values in `queueNames` https://github.com/graphile/worker/blob/ae4094600fef23a1fb053a2727762978e29aa398/website/docs/admin-functions.md#L127-L140

This was helpful. However, I think it makes sense to also warn about this on the `queueName` parameter itself, so people know _before_ there's a problem. This PR adds those warnings.

## Performance impact

N/A - docs only

## Security impact

N/A - docs only

## Checklist

- [x] N/A - docs only. I've added tests for the new feature, and `yarn test` passes.
- [x] N/A - docs only. I have detailed the new feature in the relevant documentation.
- [x] N/A - docs only. I have added this feature to 'Pending' in the `RELEASE_NOTES.md` file (if one exists).
- [x] N/A - docs only. If this is a breaking change I've explained why.

### DIFF
--- a/website/docs/library/add-job.md
+++ b/website/docs/library/add-job.md
@@ -28,21 +28,8 @@ The `addJob` arguments are as follows:
 - `payload`: an optional JSON-compatible object to give the task more context on
   what it is doing, or a list of these objects in &ldquo;batch job&rdquo; mode
 - `options`: an optional object specifying:
-
   - `queueName`: if you want certain jobs to run one at a time, add them to the
     same named queue (defaults to null which enables parallelization)
-
-    :::warning
-
-    Avoid using high cardinality values (e.g., random strings, UUIDs,
-    timestamps) for queue names as this will create many dead queues that
-    degrade performance and require
-    [periodic database cleanup](../admin-functions.md#gc_job_queues). If you
-    find yourself needing to run the `GC_JOB_QUEUES` cleanup task regularly,
-    you're likely using queue names incorrectly.
-
-    :::
-
   - `runAt`: a `Date` to schedule this task to run in the future
   - `maxAttempts`: the maximum number of attempts we'll give the job
     (Default: 25)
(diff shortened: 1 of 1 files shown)


src/interfaces.ts
```diff
@@ -84,7 +84,9 @@ export interface AddJobsJobSpec<
 
   /**
    * The queue to run this task under (only specify if you want jobs in this
-   * queue to run serially). (Default: null)
+   * queue to run serially). Avoid using high cardinality values (e.g., random
+   * strings, UUIDs, timestamps) as this degrades performance and requires
+   * periodic database cleanup. (Default: null)
```
blimmer, line 89: Feedback is welcome on the phrasing here - I don't feel strongly!

website/docs/library/add-job.md
```diff
@@ -28,8 +28,21 @@ The `addJob` arguments are as follows:
...
+
+    Avoid using high cardinality values (e.g., random strings, UUIDs,
+    timestamps) for queue names as this will create many dead queues that
+    degrade performance and require
+    [periodic database cleanup](../admin-functions.md#gc_job_queues). If you
+    find yourself needing to run the `GC_JOB_QUEUES` cleanup task regularly,
+    you're likely using queue names incorrectly.
+
+    :::
+
```
benjie, line 45: @jemgillam Please confirm this renders okay. If not, move it to below the list of options.

benjie (dismissed):
Thanks!

website/docs/library/add-job.md
benjie: The lines are all spaced out?

PR #620, opened 2026-08-06
docs(interfaces): clarify that `attempts` is 1-indexed during a run

okko:
`attempts` is incremented when a job is locked for execution, so a running handler sees `attempts === 1` on the first run and `attempts === max_attempts` on the final run. The previous wording ("How many times it has been attempted") is easy to misread as "prior failures", which leads to off-by-one final-attempt checks. Spell out the semantics on `attempts` and refine the related `last_error` comment, which read as inaccurate for the first in-flight attempt.

## Checklist

### DIFF
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -475,11 +475,17 @@ export interface DbJob {
   priority: number;
   /** When it was due to run */
   run_at: Date;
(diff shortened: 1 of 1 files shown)


benjie (approved):
Thanks!

PR #622, opened 2026-08-06
Fix localQueue running jobs from the same named queue concurrently

### LINKED ISSUE #621: `localQueue` breaks the documented serial execution guarantee for named queues
The docs state that jobs added to the same named queue run ["in series"](https://worker.graphile.org/docs/), ["one at a time"](https://worker.graphile.org/docs/library/add-job). With `worker.localQueue` enabled they run concurrently — per-queue concurrency becomes `min(localQueue.size, concurrentJobs)` instead of `1`. A single worker pool is enough to trigger it.

Nothing errors, retries or logs; every job reports success. An application relying on that guarantee to serialise read-modify-write work will silently lose writes.

Reproduced on `0.17.3` and on `main` (`cea9d60`).

## Reproduction

```bash
docker run -d --name gw-repro-pg -e POSTGRES_PASSWORD=postgres \
  -e POSTGRES_DB=gwrepro -p 54329:5432 postgres:16-alpine
npm init -y && npm i graphile-worker@0.17.3
node repro.js
```

```js
const { run, makeWorkerUtils } = require("graphile-worker");

const connectionString =
  process.env.DATABASE_URL ||
  "postgres://postgres:postgres@localhost:54329/gwrepro";
const sleep = (ms) => new Promise((r) => setTimeout(r, ms));

(async () => {
  const utils = await makeWorkerUtils({ connectionString });
  await utils.migrate();

  // A backlog of 12 jobs, all on ONE named queue.
  for (let i = 1; i <= 12; i++) {
    await utils.addJob("t", { i }, { queueName: "my-queue" });
  }

  let active = 0;
  let maxActive = 0;
  const runner = await run({
    preset: {
      worker: {
        connectionString,
        concurrentJobs: 4,
        localQueue: { size: 5 }, // <-- remove this line and maxActive is 1
      },
    },
    taskList: {
      t: async () => {
        active++;
        if (active > maxActive) maxActive = active;
        await sleep(200);
        active--;
      },
    },
  });

  while (
    (
      await utils.withPgClient((c) =>
        c.query("select 1 from graphile_worker._private_jobs limit 1"),
      )
    ).rowCount
  ) {
    await sleep(50);
  }
  await runner.stop();
  await utils.release();

  console.log(`jobs from one named queue running at once: ${maxActive}`);
  console.log(`expected per the docs: 1`);
})();
```

```
jobs from one named queue running at once: 4
expected per the docs: 1
```

Remove the one `localQueue` line and the same script prints `1`.

`size: 5` is `concurrentJobs + 1`, the value recommended in [docs/performance](https://worker.graphile.org/docs/performance). `size: 2` breaks it too; `concurrentJobs: 1` with `size: 16` does not.

## Environment

|                 |                                                                   |
| --------------- | ----------------------------------------------------------------- |
| graphile-worker | `0.17.3` (npm `latest`), and `main` @ `cea9d60` built from source |
| Node            | `v24.5.0`                                                         |
| Postgres        | `16.14`                                                           |

anavalo:
## Description

Fixes #621.

With `localQueue` enabled, jobs from the same named queue could run concurrently, violating the documented serial execution guarantee for named queues. A batch fetch (`batchSize > 1`) could lock multiple jobs belonging to the same named queue in a single statement: the queue eligibility subquery locks each available queue row once, but every job in that queue then passes the `in` membership check, so `limit batchSize` could take several of them. The local queue would hand these out to idle workers concurrently, making per-queue concurrency `min(localQueue.size, concurrentJobs)` instead of `1`. And once the first of those sibling jobs completed, `completeJobs` would unlock the queue row while the rest were still running, so other pools could lock yet more jobs from the same queue.

This PR makes batch fetches keep only the first job (by `priority`, `run_at`) per named queue and discard the rest. The discarded rows never get `locked_at`/`locked_by` set and their row locks release when the statement completes, while the queue row remains locked until the one kept job completes — restoring the invariant that a locked queue has exactly one job in flight, which `completeJobs`, `returnJobs`, and the `batchSize = 1` fetch already rely on. Jobs that aren't in a named queue are unaffected, and the `batchSize = 1` query text is byte-for-byte unchanged.

The reproduction script from #621 now reports `1` concurrent job (previously `4`), and the 12 jobs execute in insertion order.

Two new tests in `__tests__/main.runTaskList.test.ts` (the suite previously had no test exercising `localQueue`):

- jobs in the same named queue run serially even when the local queue is enabled — fails on `main`;
- jobs in different named queues run in parallel when the local queue is enabled — also fails on `main` (all four jobs start at once there), and guards against over-serialising the batch.

## Performance impact

- `localQueue` disabled (`batchSize = 1`): none — the query text is unchanged.
- Batch fetches: the added `distinct on` processes at most `batchSize` already-locked rows, which is negligible.
- Because the deduplication applies after `limit batchSize`, a fetch returns fewer than `batchSize` jobs whenever several of the selected rows share a named queue. The worst case is a backlog whose highest-precedence jobs all belong to one named queue while its tasks complete faster than `pollInterval`: each poll's fetch then selects that queue's jobs again and discards the duplicates, yielding a single job, so the pool works that queue at roughly one job per `pollInterval` and other ready jobs wait despite idle workers until the queue's backlog drains below the batch window. The effect is bounded by fetch timing: if the queue's job is still running at fetch time, its queue row is locked, its jobs are excluded, and the batch fills with jobs from other queues and unqueued jobs; a "new job" notification also triggers an immediate fetch with the same effect.
- Users of `localQueue.refetchDelay` with a non-zero `threshold` may see the refetch delay engage more often in workloads dominated by a few named queues.

## Security impact

None.

## Checklist

- [x] ~I have detailed the new feature in the relevant documentation.~ Bug fix — the documentation already describes the (serial) behaviour this PR restores.
- [x] If this is a breaking change I've explained why. — Not a breaking change: it only removes undocumented concurrency that contradicted the documented behaviour.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -46,6 +46,10 @@ Read more:
   yargs, use erasable syntax only for type-stripping support.
 - `LogLevel` export is now type only - a string union rather than a TypeScript
   const enum.
+- Fix issue where enabling `localQueue` could cause jobs from the same named
+  queue to run concurrently (violating the serial execution guarantee for named
+  queues): a single batch fetch could lock multiple jobs belonging to one named
+  queue. Batch fetches now return at most one job per named queue (#621).
 
 ## v0.17.3
 
--- a/__tests__/main.runTaskList.test.ts
+++ b/__tests__/main.runTaskList.test.ts
@@ -140,3 +140,106 @@ test("gracefulShutdown", async () =>
     const [job] = jobs;
     expect(job.last_error).toBeTruthy();
   }));
+
+test("jobs in the same named queue run serially even when the local queue is enabled", () =>
+  withPgPool(async (pgPool) => {
+    await reset(pgPool, options);
+
+    const jobPromises: Deferred[] = [];
+    try {
+      const job1: Task<"job1"> = jest.fn(() => {
+        const jobPromise = deferred();
+        jobPromises.push(jobPromise);
+        return jobPromise;
+      });
+      const tasks: TaskList = {
+        job1,
+      };
+
+      // A backlog of 5 jobs in the same named queue, before the pool starts
+      for (let i = 0; i < 5; i++) {
+        await addJob(pgPool, i);
+      }
+
+      const workerPool = runTaskList(
+        {
+          concurrency: 4,
+          preset: { worker: { localQueue: { size: 5 }, pollInterval: 10 } },
+        },
+        tasks,
+        pgPool,
+      );
+
+      for (let i = 0; i < 5; i++) {
+        await sleepUntil(() => jobPromises.length >= i + 1);
+        // Give the pool a chance to (incorrectly) hand more jobs from the
+        // same queue to the other, idle workers
+        await sleep(50);
+        expect(jobPromises).toHaveLength(i + 1);
+
+        // Complete this job, on to the next one
+        jobPromises[i].resolve();
+      }
+
+      await workerPool.gracefulShutdown();
+      await expectJobCount(pgPool, 0);
+    } finally {
+      jobPromises.forEach((p) => p.resolve());
+    }
+  }));
+
+test("jobs in different named queues run in parallel when the local queue is enabled", () =>
+  withPgPool(async (pgPool) => {
+    await reset(pgPool, options);
+
+    const started: string[] = [];
+    const jobPromisesById: Record<string, Deferred> = {};
+    try {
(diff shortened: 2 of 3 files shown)


kongallis:
Ran into this exact issue in production.

Would appreciate to see it merged when you have a moment.

Thanks @anavalo!

benjie (approved):
Looks solid; thanks!

PR #633, opened 2026-09-13
Fix explicit migration entrypoints

### LINKED ISSUE #566: Remove migrate from `getUtilsAndReleasersFromOptions`
Currently migration happens here:

https://github.com/graphile/worker/blob/91c950f7370aa2629107df1aa6a44b9f5985a1bd/src/lib.ts#L502-L505

But we're calling that from `makeWorkerUtils()`. So building a worker utils forces a migration. Kinda makes the workerUtils.migrate() method pointless!

unsady:
## Description

Fixes #566.

Removes the implicit migration from `getUtilsAndReleasersFromOptions()`, so `makeWorkerUtils()` no longer migrates the database as a side effect.

`run()`, `runOnce()`, `runMigrations()`, and the CLI continue to migrate explicitly. Added regression tests for the affected entry points.

## Performance impact

`makeWorkerUtils()` no longer performs an unnecessary migration. No change for worker startup paths.

## Security impact

None.

## Checklist

* [x] My code matches the project's code style and `yarn lint:fix` passes.
* [ ] I've added tests for the new feature, and `yarn test` passes.
* [ ] I have detailed the new feature in the relevant documentation.
* [ ] I have added this feature to 'Pending' in the `RELEASE_NOTES.md` file (if one exists).
* [ ] If this is a breaking change I've explained why.


.changeset/olive-geese-migrate.md
```diff
@@ -0,0 +1,8 @@
+---
+"graphile-worker": patch
```
benjie: If we were 1.x already, this would be a semver major change in behavior.

```suggestion
"graphile-worker": minor
```

